PR #49, opened 2022-10-27
Harden the Docker image

aymanbagabas:
* Ability to customize the VHS instance when running the evaluator.
* Check and log errors
* Refactor code
* Run the server as a different user and deescalate permissions (when supplying a GID & UID).


maaslalani (approved):

maaslalani (changes requested):
Not building on windows, I think.

aymanbagabas:
> Not building on windows, I think.

Ah, of course, no `Setuid` and `Setgid` are not a thing in Windows.

PR #54, opened 2022-10-28
feat: authorized keys

### LINKED ISSUE #52: vhs serve is currently unauthenticated
As discussed in Discord, `vhs serve` currently allows anyone to access it. 

There should likely be some authorized key check available. 

cc @caarlos0 @maaslalani

caarlos0:
closes #52

### DIFF
--- a/serve.go
+++ b/serve.go
@@ -143,7 +143,7 @@ var serveCmd = &cobra.Command{
 
 func ensureAuthorizedKeysFile(path string) error {
 	if _, err := os.Stat(path); errors.Is(err, fs.ErrNotExist) {
-		if err := os.MkdirAll(filepath.Dir(path), 0o644); err != nil {
+		if err := os.MkdirAll(filepath.Dir(path), 0o775); err != nil {
 			return fmt.Errorf("could not create %s: %w", path, err)
 		}
 		if err := os.WriteFile(path, nil, 0o600); err != nil {
--- a/serve.go
+++ b/serve.go
@@ -26,12 +26,11 @@ const (
 )
 
 type config struct {
-	Port int    `env:"PORT" envDefault:"1976"`
-	Host string `env:"HOST" envDefault:""`
-	Key  string `env:"KEY" envDefault:""`
-	GID  int    `env:"GID" envDefault:"0"`
-	UID  int    `env:"UID" envDefault:"0"`
-
+	Port               int    `env:"PORT" envDefault:"1976"`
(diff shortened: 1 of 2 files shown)


maaslalani (approved):
Thank you @caarlos0 !

maaslalani:
We'll also probably want to add some docs to the VHS server section

serve.go
```diff
@@ -31,6 +33,8 @@ type config struct {
 	Key  string `env:"VHS_KEY" envDefault:""`
 	GID  int    `env:"VHS_GID" envDefault:"0"`
 	UID  int    `env:"VHS_UID" envDefault:"0"`
+
+	AuthorizedKeysPath string `env:"VHS_AUTHORIZED_KEYS_PATH" envDefault:"${HOME}/.ssh/authorized_keys" envExpand:"true"`
```
aymanbagabas, line 37: Is this default going to be problematic for Windows users?

caarlos0: good question... and I don't really know... what I did is to create the filepath if it does not exist, so at worst it will use an empty file... maybe we should change it to default empty though?

maaslalani: Also, what if someone wants to keep a publicly accessible server?

jolheiser: > Is this default going to be problematic for Windows users?

Assuming `${HOME}` is using the `os.Expand` functionality behind the scenes, technically Windows would use `USERPROFILE` and so this might not work as intended.
If possible, `os.UserHomeDir` would probably be preferable.

----

Perhaps to tackle both concerns with one stone, this could be optional entirely. If set, the user will be able to describe their home path as needed, and if not set then there is no auth.

serve.go
```diff
@@ -26,19 +26,22 @@ const (
 )
 
 type config struct {
-	Port int    `env:"VHS_PORT" envDefault:"1976"`
-	Host string `env:"VHS_HOST" envDefault:""`
-	Key  string `env:"VHS_KEY" envDefault:""`
-	GID  int    `env:"VHS_GID" envDefault:"0"`
-	UID  int    `env:"VHS_UID" envDefault:"0"`
+	Port               int    `env:"PORT" envDefault:"1976"`
+	Host               string `env:"HOST" envDefault:""`
```
caarlos0, line 30: fwiw this should probably default to localhost only... otherwise the server is open by default with no auth...

aymanbagabas (approved):

PR #55, opened 2022-10-28
feat: Set Shell

### LINKED ISSUE #44: issue: `vhs` thinking that I am running on `WSL` when it is not the case
I have downloaded `vhs` from [GitHub Release](https://github.com/charmbracelet/vhs/releases/download/v0.1.0/vhs_0.1.0_Windows_x86_64.zip).  I am trying to make some GIF for my examples of the project [box-cli-maker](https://github.com/Delta456/box-cli-maker) with the following code:

```
Output demo.gif

Set FontSize 46
Set Width 1200
Set Height 600

# Type a command in the terminal.
Type "go run path\git\box-cli-maker\examples\unicode.go"

# Pause for dramatic effect...
Sleep 500ms

# Run the command by pressing enter.
Enter

# Admire the output for a bit.
Sleep 5s
```

The GIF successfully produces but it throws me an error as it expects `go` to be installed on **Ubuntu** when I am not running `vhs` on WSL.

demo

It seems weird to me how this happened 🤔

### LINKED ISSUE #47: Support shell different than bash

caarlos0:
no idea if this is the best way to do this, just wanted to play with it a bit :D 

also check treesitter patch @ https://github.com/charmbracelet/tree-sitter-vhs/pull/1

closes #47
closes #44


vhs.go
```diff
@@ -98,8 +101,15 @@ func (vhs *VHS) Setup() {
 	vhs.CursorCanvas, _ = vhs.Page.Element("canvas.xterm-cursor-layer")
 
 	// Set Prompt
+	shellSetup := ""
+	switch vhs.Options.Shell {
+	case "", "bash":
+		shellSetup = fmt.Sprintf(` set +o history; unset PROMPT_COMMAND; export PS1="%s"; clear;`, vhs.Options.Prompt)
+	default:
```
caarlos0, line 108: here I *think* we should probably add switches for the most common shells, at least zsh and fish, I would say

caarlos0:
> Nice implementation @caarlos0! I was originally thinking we would have to start `ttyd` with the correct shell but I think this is a better approach.

I thought about it too... but I think then we will have to handle dependencies too... or just not handle the shell dependency... I'm not sure which is better, happy to hear what you think, and can change either way

maaslalani:
> I thought about it too... but I think then we will have to handle dependencies too... or just not handle the shell dependency... I'm not sure which is better, happy to hear what you think, and can change either way

Let's leave it like this for now. We can always change the implementation later because the  API / language will stay the same.

vhs.go
maaslalani: Agreed! This is great!

Delta456:
For adding Powershell 5 and Powershell 7 support you have to write `powershell` and `pwsh` respectively.

dlvhdr:
Thanks for this! With this I'll be able to create demo gifs for my dotfiles repo <3

caarlos0:
> For adding Powershell 5 and Powershell 7 support you have to write `powershell` and `pwsh` respectively.
> 
> **EDIT**: For opening Command Prompt you need to write `cmd`.

I don't have a windows machine, so I don't know how any of that works... 

One thing we could do is, instead of defaulting to `%s --login`, default to `%s` only, so you can set your shell to whatever you need... and we have some sensible defaults for bash, zsh and fish which are arguably the most commonly used ones... 

Another advantage of that strategy is that, if you want to record using your dotfiles, as @dlvhdr mentioned, you can set the shell to `myshell --login` and vhs will not do anything with it.

wdyt?

Delta456:
> > For adding Powershell 5 and Powershell 7 support you have to write `powershell` and `pwsh` respectively.
> > **EDIT**: For opening Command Prompt you need to write `cmd`.
> 
> I don't have a windows machine, so I don't know how any of that works...
> 
> One thing we could do is, instead of defaulting to `%s --login`, default to `%s` only, so you can set your shell to whatever you need... and we have some sensible defaults for bash, zsh and fish which are arguably the most commonly used ones...
> 
> Another advantage of that strategy is that, if you want to record using your dotfiles, as @dlvhdr mentioned, you can set the shell to `myshell --login` and vhs will not do anything with it.
> 
> wdyt?

I can help with the Windows functionality.

michaeltlombardi:
Dropping the `--login` would certainly help for PowerShell, Windows PowerShell, and the command prompt. More generally, being able to specify args for the shell would be neat.

FWIW wrt testing, PowerShell (`pwsh`) is cross-platform and should be testable locally without a Windows machine.

I write docs for PowerShell, happy to take a poke, answer questions, and help out with whatever.

michaeltlombardi:
@caarlos0 I didn't notice anything in the PowerShell code you wrote that would do that, wondering if it's strange behavior elsewhere. Pulling down your branch locally to try it in WSL (and Windows, if we get WSL happy), will let you know what I discover.

griimick:
Hey! I was dealing with a similar issue this afternoon, the extra `e` is what you see when you start typing before letting the last executing setup commands finish. I see a lot more characters when I don't set `Sleep` after `Set Shell`

If I set a sleep of 1 sec, I see this

pwsh

and it goes away if I set Sleep for 2sec,

pwsh

michaeltlombardi:
Glad you got things working, @griimick - I've hit snag after snag 😓

examples/shell/pwsh.tape
```diff
@@ -0,0 +1,8 @@
+Output examples/shell/pwsh.gif
+Output examples/shell/pwsh.ascii
+
+Set Shell "pwsh"
+
+Type 'echo "current shell: $ShellId"'
```
michaeltlombardi, line 6: For PowerShell, it's more idiomatic to either write the string (without `echo` beforehand), or use `Write-Output` here. I would lean towards the latter for better optics with PowerShell users.

maaslalani:
@caarlos0 simplified the logic a little bit, let me know if it looks good to you.

prompt.go
```diff
@@ -1,79 +1,40 @@
...
-	fish    struct{}
-	pwsh    struct{}
-	generic struct{}
+// Supported shells of VHS
+const (
+	BASH       = "bash"
+	CMD        = "cmd"
+	FISH       = "fish"
+	POWERSHELL = "pwsh"
+	ZSH        = "zsh"
```
caarlos0, line 9: no need to be uppercase

examples/settings/set-shell.tape
```diff
@@ -0,0 +1,17 @@
...
+# Set Shell zsh
+# Set Shell bash
+# Set Shell pwsh
+
+Sleep 1s
+Type "I"
+Sleep 1s
+Type " am using the friendly interactive shell"
+Sleep 1s
+Enter
```
caarlos0, line 16: this enter will cause the previous type to error

prompt.go
```diff
@@ -1,79 +1,40 @@
...
+		Prompt:  `function fish_prompt; echo -e "$(set_color 5B56E0)> $(set_color normal)"; end`,
+		Command: `clear; fish --login --private -C 'function fish_greeting; end' -C '%s'`,
+	},
+	POWERSHELL: {
+		Prompt:  "Function prompt {Write-Host \"> \" -ForegroundColor Blue -NoNewLine; return \"`0\" }",
+		Command: ` clear; pwsh -Login -NoLogo -NoExit -Command 'Set-PSReadLineOption -HistorySaveStyle SaveNothing; %s'`,
+	},
+	CMD: {
+		Prompt:  "> ",
+		Command: ` clear; export PS1="%s"; clear`,
```
caarlos0, line 38: is this right? does CMD use `cls` instead of `clear`? also not sure about `export PS1` for cmd

caarlos0:
@maaslalani 
I made the shell interface because I'm not sure how other shells would behave and if just running the 2 command will be enough... but I think we can always refactor later if needed, so, lgtm!

caarlos0:
I also liked the tape files to verify that it is indeed working as expected... its kinda easy to break it 🤔

prompt.go
griimick: ` cls; set prompt=$E[34m$g$E[0m` will work as expected in `cmd` ([demo](https://gist.github.com/griimick/d3d47c7f890388d7ab018cfb598b9102)).

caarlos0: if I want to set to `>`, can I simply `cls; set prompt=">"`:

caarlos0:
> @caarlos0 simplified the logic a little bit, let me know if it looks good to you.

oh, and it also breaks using a custom shell, e.g. if someone wants to set it to `elvish`, or some custom bash... not sure how relevant that it, but anyway...

maaslalani:
> > @caarlos0 simplified the logic a little bit, let me know if it looks good to you.
> 
> oh, and it also breaks using a custom shell, e.g. if someone wants to set it to `elvish`, or some custom bash... not sure how relevant that it, but anyway...

👍 I can fix this one

PR #67, opened 2022-10-28
feat: (fixes #14) add Require command

### LINKED ISSUE #14: Fail early when missing required programs
VHS depends on the host computer to execute terminal commands (including installed software).
It is possible for a tape that runs on one computer (with a program installed) will generate incorrectly on a different computer (with the program missing).

For example, the following tape:

```elixir
Output glow.gif

Type "glow"
Enter
Sleep 5
```

requires that `glow` is installed on the machine that VHS is running on.
On a machine that doesn't have `glow` installed the recording will still succeed but the output will show "bash: command not found: glow".

There are a few different solutions we can employ:

1. Introduce a `Require` command

The `Require` command would act as a pre-check to ensure the computer running VHS has the correct dependencies for the `.tape` file in the path. For example:

```elixir
Output glow.gif

Require glow,foo,bar

Type "glow"
Enter
Sleep 5
```

Now, when running VHS the running instance will exit as soon as it realizes that `glow` is not installed on the machine. More importantly, it will not override the existing `glow.gif` if it exists. The downside is that the tape writer must enumerate the dependencies at the top of the file manually (which might not be a big deal).

2. Introduce a `set -e`-like option 

> The `set -e` option instructs bash to immediately exit if any command [1] has a non-zero exit status.

We can introduce a command / setting that tells VHS to stop recording if any of the commands has a non-zero exit code, as would be in the case of a program not being installed.

```elixir
Output glow.gif

Set Exit ... # OnFailure

Type "glow"
Enter
Sleep 5
```

This tape would exit after `Type "glow"` + `Enter` commands, when it realizes that the command failed. The downside of this is that sometimes demos may involve exiting with non-zero status codes to be part of the GIF.

3. Introduce an `Exec` command

```elixir
Output glow.gif

Exec "glow"

Sleep 5
```

This `Exec` command would act as a `Require <program>` + `Type <program">` + `Enter`. The downside of this approach is that it is harder to modify aspects of the typing speed and delay between typing the last letter and pressing enter, though these can be solved with something such as `Exec@200ms "glow" 1s` (but maybe that is a bit complicated syntax)

abhijit-hota:
As per the discussion in #14 and the verdict in the comment: https://github.com/charmbracelet/vhs/issues/14#issuecomment-1281083943, this PR aims to implement the `Require` method that detects missing binaries to fail early.


parser.go
```diff
@@ -229,6 +231,36 @@ func (p *Parser) parseHide() Command {
...
+		}
+		p.nextToken()
+
+		cmd.Args += p.cur.Literal
+	}
+
+	// check for trailing commas
+	if p.cur.Type == COMMA {
+		p.errors = append(p.errors, NewError(p.cur, "Stray comma found in Require"))
+	}
```
abhijit-hota, line 259: The error messages for these can be improved.

maaslalani:
Nice! @abhijit-hota This is great!

command.go
```diff
@@ -138,6 +142,15 @@ func ExecuteHide(c Command, v *VHS) {
 	v.PauseRecording()
 }
 
+func ExecuteRequire(c Command, v *VHS) {
+	for _, t := range strings.Split(c.Args, ",") {
+		if _, err := exec.LookPath(t); err != nil {
+			log.SetFlags(0)
+			log.Fatalf("The VHS tape requires the following binaries:\n%s\n", c.Args)
```
caarlos0, line 149: I think it would be bit more idiomatic to do:

```suggestion
			fmt.Printf("The VHS tape requires the following binaries:\n%s\n", c.Args)
			os.Exit(1)
```

caarlos0: also, probably worth mentioning which of the required binaries is missing

abhijit-hota: Thanks for the review.

> ```go
> fmt.Printf("The VHS tape requires the following binaries:\n%s\n", c.Args)
> ```

I mainly chose `log.Fatalf` because it sends to `os.Stderr` as well as `exit`s with 1.

Should I go with:

```go
fmt.Fprintf(os.Stderr, "The VHS tape requires the following binaries:\n%s\n", c.Args)
os.Exit(1)
```

> worth mentioning which of the required binaries is missing

That's a good idea! Will implement.

token.go
```diff
@@ -16,6 +16,7 @@ const (
 	AT             = "@"
 	EQUAL          = "="
 	PLUS           = "+"
+	COMMA          = ","
```
maaslalani, line 19: Just thinking out loud for the syntax, these are some other options we have.

One program per require (a bit verbose):
```
Require foo
Require bar
Require baz
```

Whitespace separated (a bit tricky)
```
Require foo bar baz
```

Colon-separated (resembles `$PATH`)
```
Require foo:bar:baz
```

Comma separated (current impl):
```
Require foo,bar,baz
```

abhijit-hota: The comma separated is the most unambiguous, personally.

caarlos0: I would prefer the one require per line, actually, otherwise if you have a lot of requirements the line width would get too big.

also, one per line make code reviews easier, as its either adding or remove one line, instead of changing a big one.

abhijit-hota: One require per line makes sense. They are like `import`s in other programming languages.

Just pitching another one:
```sh
Require foo

# or

Require (
    foo
    bar
    neofetch
    fdfind
    btop
)
```
like Golang `import`s.

Granted, it'll be tougher to parse.

maaslalani: @abhijit-hota I think we can start with one per line and each needs a require since that more closely matches how the rest of the language works i.e. more command based.

Also, if people really want to keep everything on one line they can do:

```
Require foo Require bar Require baz
```

So, I think this is a decent solution!

abhijit-hota: Makes sense. Will implement this!

abhijit-hota: On second thought, if we implement this syntax, we cannot check for all binaries that are missing altogether. 

As we have to check the `Require` one by one.

maaslalani: One way around this is to make `ExecuteRequire` just add the programs to a list on VHS and then when you hit a non-require command evaluate all the requires in that group and clear the list.

maaslalani: So you basically group requires:

```
# Will exit early if foo, bar, baz are missing
Require foo
Require bar
Require baz

Type "foo bar baz"
Enter

# Will exit if qux is missing
Require qux
Type "qux"
```

maaslalani: Another idea is to do a pre-run of all the commands and filter all the require ones, add them to a list and exit if any are missing.

abhijit-hota: So to summarise, we have the following 2 behaviours we need to decide on. Please correct me if I'm wrong or missing anything.

### Behaviour 1
- All the `Require`s are checked before the actual recording is started.
- Outputs all the missing binaries altogether, if any.
- Note on semantics:
    - If we intend the command to be used anywhere in the code, we have to pre-parse the whole tape. 
    - Another alternative is to an only allow it at the start of a `tape`. 

### Behaviour 2
- Each `Require` is checked individually.
- Outputs only the binary at which it failed.
- Can be used anywhere.

maaslalani: Yep! I'm actually okay with either behaviour. Also happy to set a requirement that they are at the top of the file if they want to be used, we have a similar requirement of Output and Set commands.

maaslalani: Hmm, I believe its only settings which must be at the top then, if there is a setting in the middle of a tape it will be ignored

maaslalani: Imo we can allow requires to be anywhere and exit at the point they are which is a more expected behaviour

abhijit-hota: The following piece of code only checks for Set commands and that's the reason Output isn't ignored wherever is it.

https://github.com/charmbracelet/vhs/blob/1a2716c7684d409ed575a9fd5932c3c2086d711a/evaluator.go#L92-L98

abhijit-hota: @maaslalani The problem I think with executing `Require`s individually is this:

```sh
Require foo
Type foo
Enter

Sleep 15s

Require bar
Type bar
Enter
```

Now the user will not know about the `bar` dependency for like ≈20s. The original issue was to "fail-first" and I think we should do that as soon as possible.

abhijit-hota: I will implement the pre-parsing method and get back to you.

maaslalani: Cool, thanks! Yeah I believe we should pre-parse and always encourage (in our examples) to put all the `Require` statements at the top of the file as a best practice.

abhijit-hota: Just implemented it. Take a look. @maaslalani

command.go
```diff
@@ -138,6 +142,28 @@ func ExecuteHide(c Command, v *VHS) {
 	v.PauseRecording()
 }
 
+// ExecuteRequire is a CommandFunc that checks if all the binaries mentioned in the
+// Require command are present. If not, it exits with a non-zero error.
+func ExecuteRequire(c Command, v *VHS) {
+	unavailableBinaries := make([]string, 0)
+
+	requiredBinaries := strings.Split(c.Args, ",")
```
abhijit-hota, line 150: I don't feel very good about this function as it's not inline with the actual `Require` syntax. 

Should we just move this logic to the `checkRequires` function?

evaluator.go
```diff
@@ -98,3 +107,18 @@ func Evaluate(tape string, out io.Writer, opts ...EvaluatorOption) error {
...
+		if cmd.Type == REQUIRE {
+			fmt.Fprintln(out, cmd.Highlight(false))
+			allBinaries = append(allBinaries, cmd.Args)
+		}
+	}
+	requireCmd := Command{
+		Type: REQUIRE,
+		Args: strings.Join(allBinaries, ","),
+	}
+	requireCmd.Execute(&v)
```
abhijit-hota, line 123: Instead of deceptive calling the Require command like this, can we implement the binary checking logic directly here? (as I commented earlier).

We won't need an `ExecRequire` function then.

maaslalani:
@abhijit-hota Can I get push access to your fork. I've made some changes that (I think) fix all our concerns.

maaslalani:
Added an errors field for runtime errors on the VHS struct:

image


That way the `ExecuteRequire` can just append to this errors stuff and not exit right away, allowing us to collect all the errors.

evaluator.go
maaslalani: I've made it so that `ExecRequire` only looks at it's args and appends an error to the VHS struct.

command.go
maaslalani: I agree, I've changed it so that all the logic is more maintainable by introducing runtime errors on the VHS struct.

abhijit-hota:
Have invited you as collaborator. @maaslalani

maaslalani:
@abhijit-hota Thanks, pushed up changes.

I think this makes it a lot more like the other commands and now allows us to have runtime errors as well.

It does mean that `Require`s must always be near the top of the file but I think that's important for readability anyway so that people looking at tape files know exactly what programs they need.

abhijit-hota:
Great! :tada: 

> It does mean that Requires must always be near the top of the file but I think that's important for readability anyway so that people looking at tape files know exactly what programs they need.

Got it. We should document this in the README as well.

maaslalani (approved):
Thanks for the great work! @abhijit-hota

PR #74, opened 2022-10-28
docs: add server settings

aymanbagabas:
Adds vhs server settings

### DIFF
--- a/README.md
+++ b/README.md
@@ -137,19 +137,21 @@ applications on the host so you don't need to install them on your machine.
 
 You can customize the VHS server by setting the following environment variables:
 
-* `VHS_PORT` The port to listen on. Defaults to `1976`.
-* `VHS_HOST` The host to listen on. Defaults to `localhost`.
-* `VHS_GID` The GID to run the server as. Defaults to the current user's GID.
-* `VHS_UID` The UID to run the server as. Defaults to the current user's UID.
-* `VHS_KEY_PATH` The SSH key to use. Defaults to `.ssh/vhs_ed25519`.
-* `VHS_AUTHORIZED_KEYS_PATH` The authorized keys file to use. Disabled by default.
+| Environment                | Description                        | Default             |
+| -------------------------- | ---------------------------------- | ------------------- |
+| `VHS_PORT`                 | The port to listen on.             | `1976`              |
+| `VHS_HOST`                 | The host to listen on.             | `localhost`         |
+| `VHS_GID`                  | Group ID to run the server as.     | Current user's GID  |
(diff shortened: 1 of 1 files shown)


caarlos0:
@maaslalani I like the idea, then we can go further into detail if needed be

README.md
```diff
@@ -128,12 +128,31 @@ go install github.com/charmbracelet/vhs@latest
...
+VHS has an SSH server built in! When you self-host VHS you can access it as
 though it were installed locally. VHS will have access to commands and
 applications on the host so you don't need to install them on your machine.
 
-Ensure you've set `~/.ssh/authorized_keys` with the public keys to which
-you want grant SSH access. You can change the file VHS looks at by setting
+<details>
+<summary>
+
+### Server Configuration
```
caarlos0, line 138: I think the title looks a bit weird inside a summary... would probably change it to 

```html
<summary>Server Configuration</summary>
``` 

only

caarlos0 (approved):

README.md
```diff
@@ -128,13 +128,30 @@ go install github.com/charmbracelet/vhs@latest
...
+| `VHS_PORT`                 | The port to listen on.             | `1976`              |
+| `VHS_HOST`                 | The host to listen on.             | `localhost`         |
+| `VHS_GID`                  | Group ID to run the server as.     | Current user's GID  |
+| `VHS_UID`                  | User ID to run the server as.      | Current user's UID  |
+| `VHS_KEY_PATH`             | The path to the SSH key to use     | `.ssh/vhs_ed25519`  |
+| `VHS_AUTHORIZED_KEYS_PATH` | The path to authorized keys files  | Empty               |
+
+> **Warning**:
+> An empty `VHS_AUTHORIZED_KEYS_PATH` means that your server is publically
+> accessible. Change this to `~/.ssh/authorized_keys` or another path to ensure
```
aymanbagabas, line 151: I think this should be `.ssh/authorized_keys` instead to create the file in the binary's CWD

caarlos0 (approved):

README.md
caarlos0: it can be, but i'll not create the file if it doesn't exist...

I openend an issue on wish to work on auto-reloading the authorized keys file, maybe we could handle it there too

README.md
```diff
@@ -128,13 +128,30 @@ go install github.com/charmbracelet/vhs@latest
 
 ## The VHS Server
 
-VHS has an SSH server built in! When you self host VHS you can access it as
+VHS has an SSH server built in! When you self-host VHS you can access it as
```
meowgorithm, line 131: In this case "self host" actually doesn't need a hyphen because it's not modifying a noun.

PR #83, opened 2022-10-28
feat: support named themes

### LINKED ISSUE #78: feat: set theme
Discussed in Discord

i think you should be able to do `Set Theme TokyoNight`
- [ ] include supported themes in the README

caarlos0:
Initial implementation, let me know what you think!

closes #78 
refs https://github.com/catppuccin/catppuccin/issues/1711

Thanks to:
- @jolheiser catppuccin colors
- @dlvhdr tokyo night theme

--- 

EDITS:

- ended up importing themes from https://github.com/atomcorp/themes
- `go generate` will sync the json theme files, and generate the `THEMES.md` file
- added a `vhs themes` so users can `vhs themes | grep -i "catpp"` et al to find the correct name quickly
- fixed some docs instances which were using wrong field names
- patched treesitter @ https://github.com/charmbracelet/tree-sitter-vhs/pull/2


maaslalani:
@caarlos0 do you have `git-lfs` installed?

maaslalani:
I see that the GIFs are added directly to the git repo, without uploading to `git-lfs`

caarlos0:
oh, I dont, will install @maaslalani

aymanbagabas:
@caarlos0 you might need to rebase this pr

command.go
```diff
@@ -259,13 +261,22 @@ func ExecuteSetLineHeight(c Command, v *VHS) {
...
+			v.Options.Theme = DefaultTheme
+			return
+		}
+	} else {
+		if theme, ok := Themes()[c.Args]; ok {
+			v.Options.Theme = theme
+		} else {
+			v.Options.Theme = DefaultTheme
+			return
+		}
```
maaslalani, line 276: Let's check if the theme is in the `Themes` map first (it won't be if given JSON) and then try parsing JSON and then return if error. I think that should clean up the error handling and if statements a bit

theme.go
```diff
@@ -13,12 +14,130 @@ import (
 	"github.com/charmbracelet/glamour/ansi"
 )
 
+// Themes return the list of themes.
+func Themes() map[string]Theme {
```
maaslalani, line 18: How come this is a function rather than a `var`?

caarlos0:
@aymanbagabas wont the squash and merge do that?

theme.go
```diff
@@ -13,12 +14,130 @@ import (
...
+	return map[string]Theme{
+		"TokyoNight": {
```
dlvhdr, line 20: This is tokyo night storm... Apparently there are variations 🥲

theme.go
caarlos0: just so its modifiable... but on second thought that might not be worth it

themes.go
```diff
@@ -0,0 +1,115 @@
+package main
+
+// Themes return the list of themes.
+var Themes = map[string]Theme{
+	"TokyoNight-Storm": {
```
maaslalani, line 5: @caarlos0 Thinking about this a bit more, I think it would be a little bit of a pain to store all these themes and maintain this list. This project seems to have all of the themes stored as JSON in a format very similar to what we want.

https://windowsterminalthemes.dev

I wonder if we can use this to piggyback off of to have a ton of themes without needing to maintain all of them ourselves. We could also standardize on a different base16 format (like alacritty color themes) and have all of them be generated or fetched from that.

caarlos0: good idea, no cattpuccin theme there though, will pr it and change here


https://github.com/atomcorp/themes/blob/master/themes.json 

we can probably embed this file

maaslalani: Nice, yeah sounds like a plan! Thank you!

maaslalani: Also regarding theme names I think spaces are the way to go (rather than dashes):

```
Set Theme "Catpuccin Mocha"
```

caarlos0: https://github.com/atomcorp/themes/pull/86

caarlos0: @maaslalani just pushed a first version of all that, catppuccin is not there yet thougn, need to wait for them to merge

maaslalani: Thank you so much @caarlos0, this is incredible! This is going to be much more maintainable.

caarlos0: totally yes

caarlos0: thank you for the idea of that repo, didnt know it :D

README.md
```diff
@@ -289,11 +289,19 @@ Set the theme of the terminal with the `Set Theme` command. The theme value
...
 
 <img alt="Example of changing the theme to Whimsy" src="https://stuff.charm.sh/vhs/examples/theme.gif" width="600" />
 
+You can also set some themes by their name:
+
+- `Set Theme Catppuccin-Frappe`
+- `Set Theme Catppuccin-Latte`
+- `Set Theme Catppuccin-Macchiato`
+- `Set Theme Catppuccin-Mocha`
+- `Set Theme TokyoNight-Storm`
```
caarlos0, line 303: wil have to update this... or maybe we just point to the website? Or do we add a command to list all themes?

caarlos0:
what do you think: do we add a `vhs themes` command, or keep it in the man page, or both?

jolheiser:
> what do you think: do we add a `vhs themes` command, or keep it in the man page, or both?

Maybe both? Alternatively, man pages and a markdown file?

As well, maybe even in a follow up, could the list be generated in go code so that vhs can verify tape files include a valid theme (if not json)?

caarlos0:
> Maybe both? Alternatively, man pages and a markdown file?

yup, done!

sync-themes.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/sh
```
maaslalani, line 1: Can we put this in a Makefile, so that way we can just `make themes.json` after deleting it:

```
themes.json:
   ....
```

What do you think? If you prefer keeping a binary let's put this in a `bin/` folder.

themes.go
```diff
@@ -0,0 +1,98 @@
...
+					Red:           theme.Red,
+					BrightRed:     theme.BrightRed,
+					Green:         theme.Green,
+					BrightGreen:   theme.BrightGreen,
+					Yellow:        theme.Yellow,
+					BrightYellow:  theme.BrightYellow,
+					Blue:          theme.Blue,
+					BrightBlue:    theme.BrightBlue,
+					Magenta:       theme.Purple,
+					BrightMagenta: theme.BrightPurple,
```
maaslalani, line 63: If you want, we can change `magenta` and `brightMagenta` to be `purple` / `brightPurple`, so that we can be consistent. It wouldn't be a breaking change but it would make purples of the themes be off.

maaslalani: Alternatively we could `sed s/purple/magenta` (when syncing) in the `themes.json` file so that we don't have to do this conversion / renaming in code.

maaslalani (approved):

maaslalani:
Looking _really_ good @caarlos0, thank you for this!

themes.go
caarlos0: I did that initially, but seems like xterm uses magenta instead of purple... there are other differences too, like cursor color and selection background.

sync-themes.sh
caarlos0: I can move into a bin or scripts folder, and also add a makefile... 

right now, you can also run it with `go generate` fwiw

maaslalani: Ah if it's available with `go generate` let's delete this file then!

maaslalani: (personally, think it's good to have only one way of doing things)

command_test.go
```diff
@@ -15,3 +19,70 @@ func TestCommand(t *testing.T) {
 		t.Errorf("Expected %d commands, got %d", numberOfCommands, len(CommandTypes))
 	}
 }
+
+func TestExecuteSetTheme(t *testing.T) {
+	browser := rod.New().MustConnect()
```
maaslalani, line 24: These tests seem a bit heavy for `go test` (i.e. don't want to need a browser installed to run tests), we already get integration testing by generating the GIFs. Can we just run `ApplyTheme` to check the go struct is updated?

command.go
```diff
@@ -269,13 +271,14 @@ func ExecuteSetLineHeight(c Command, v *VHS) {
 
 // ExecuteSetTheme applies the theme on the vhs.
 func ExecuteSetTheme(c Command, v *VHS) {
-	err := json.Unmarshal([]byte(c.Args), &v.Options.Theme)
+	var err error
+	v.Options.Theme, err = getTheme(c.Args)
```
maaslalani, line 275: The new `vhs.Errors` might be useful here. We should probably fail fast and not generate the full GIF if the theme will be wrong.

themes.go
```diff
@@ -0,0 +1,98 @@
...
+// Themes return the list of themes.
+func Themes() map[string]Theme {
+	themesOnce.Do(func() {
+		var themes []windowsTerminalTheme
+		for _, bts := range [][]byte{themesBts, customThemesBts} {
+			if err := json.Unmarshal(bts, &themes); err != nil {
+				fmt.Fprintf(os.Stderr, "could not load themes.json: %s\n", err)
+				os.Exit(1)
+			}
+			for _, theme := range themes {
```
maaslalani, line 45: How come we need to do this for all the themes?
Each tape file can use only one theme. It feels like a lot of memory going to be used to do this conversion for only using one theme.
Perhaps we can keep all the themes just how they are (with purple and stuff) and then convert when we want to use it? Just an idea, though.

sync-themes.sh
caarlos0: done

themes.go
caarlos0: done

themes.go
caarlos0: One thought I had was to actually filter and load just the one theme asked... which seems more reasonable actually, will do that I think

main.go
```diff
@@ -68,6 +67,27 @@ var (
...
+				}
+				return
+			}
+
+			fmt.Fprintln(cmd.OutOrStdout(), "# Themes")
+			fmt.Fprintln(cmd.OutOrStdout())
+			for _, theme := range sortedThemeNames() {
+				fmt.Fprintf(cmd.OutOrStdout(), "* `%s`\n", theme)
+			}
+		},
```
maaslalani, line 88:
```suggestion
            var prefix, suffix string
			if markdown {
			    fmt.Fprintln(cmd.OutOrStdout(), "# Themes\n")
			    prefix, suffix = "* `", "`"
			}
			for _, theme := range sortedThemeNames() {
				fmt.Fprintf(cmd.OutOrStdout(), "%s%s%s\n", prefix, theme, suffix)
			}
		},
```

maaslalani: Not really sure if this is cleaner / more legible. Just another approach.

maaslalani (approved):

themes.go
maaslalani: Sounds great! Thank you

examples/settings/set-theme-name.tape
```diff
@@ -0,0 +1,7 @@
+Output examples/settings/theme-name.gif
+Set FontSize 42
+Set Height 225
+Set Theme "Catppuccin Latte"
+
+Type "I can't make up my mind."
+Sleep 1
```
maaslalani, line 7: This one seems to be ending a bit quickly 🤔

maaslalani:
Looking good @caarlos0, will merge this soon!

examples/settings/set-theme-name.tape
caarlos0: fixed, I think I might have press ctrl+c while processing or something

maaslalani:
Thanks so much @caarlos0, incredible work. Thanks for baring with me with the code review ❤️

caarlos0:
was very fun to work on it, thanks for reviewing & merging @maaslalani 💙

PR #84, opened 2022-10-28
fix: bump minimum required Go version to 1.18

muesli:
The minimum required Go version to build vhs is 1.18.


maaslalani:
Thanks @muesli ❤️

PR #90, opened 2022-10-29
fix: handle newline on crlf

### LINKED ISSUE #80: Wrong number of lines in case of CRLF files
The `isNewLine` function below checks for `\r` and `\n` separately. 

https://github.com/charmbracelet/vhs/blob/ab65901d8f2301e7585f482b2514b62a62cf86e8/lexer.go#L204-L207

The CRLF line ending is `\r\n` and hence, the line count doubles as they are **counted as two different newlines** here:

https://github.com/charmbracelet/vhs/blob/ab65901d8f2301e7585f482b2514b62a62cf86e8/lexer.go#L151-L162

This messes up error logs and sometimes gives out of range errors due to this code here:

https://github.com/charmbracelet/vhs/blob/1a3529697527c3f5f2563c8d8a097c45491d2880/main.go#L116

caarlos0:
closes #80

### DIFF
--- a/lexer.go
+++ b/lexer.go
@@ -29,7 +29,7 @@ func (l *Lexer) readChar() {
 func (l *Lexer) NextToken() Token {
 	l.skipWhitespace()
 
-	var tok = Token{Line: l.line, Column: l.column}
+	tok := Token{Line: l.line, Column: l.column}
 
 	switch l.ch {
 	case 0:
@@ -93,7 +93,7 @@ func (l *Lexer) readComment() string {
 	pos := l.pos + 1
 	for {
 		l.readChar()
-		if isNewLine(l.ch) || l.ch == 0 {
+		if isNewLine([2]byte{l.ch, l.peekChar()}) || l.ch == 0 {
 			break
 		}
 	}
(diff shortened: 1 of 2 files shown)


lexer.go
```diff
@@ -202,8 +202,8 @@ func isWhitespace(ch byte) bool {
 }
 
 // isNewLine returns whether a character is a newline.
-func isNewLine(ch byte) bool {
-	return ch == '\n' || ch == '\r'
+func isNewLine(ch [2]byte) bool {
```
abhijit-hota, line 205: What do you think about passing the lexer directly to this function? The same pattern is being used every time we call it:

```go
isNewLine([2]byte{l.ch, l.peekChar()})
```

The comment says `character` but it's actually not.

maaslalani:
Another way we could possibly do this is to have a `isCarriageReturn` function and always skip that no matter what.

abhijit-hota:
We  can also retain the original function (with a comment noting the catch) but while incrementing the line numbers, we only check `\n` like mentioned below and not use yhe function.

```go
if l.ch == '\n' {
    l.lines++
}
```

PR #92, opened 2022-10-29
fix: require ttyd 1.7.2

muesli:
If I'm not mistaken, we require ttyd 1.7.2?


aymanbagabas (approved):

maaslalani:
Yes, my mistake! Thanks for catching this Muesli!

PR #99, opened 2022-10-30
chore: use a context to control the recording process

muesli:
This prevents the render goroutine from directly printing to stdout and ensures a controlled tear-down.

### DIFF
--- a/evaluator.go
+++ b/evaluator.go
@@ -1,9 +1,11 @@
 package main
 
 import (
+	"context"
 	"errors"
 	"fmt"
 	"io"
+	"log"
 	"strings"
 )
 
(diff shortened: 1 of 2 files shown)


caarlos0 (approved):
looks good!

as a nitpick: could probably use exec with context when starting ttyd and ffmpeg as well.

muesli:
@caarlos0 coming up in a separate PR!

maaslalani:
Thanks @muesli!

On a semi-related note, would we be able to make the screenshot capturing stuff faster using goroutines? I tried playing around with it a bit myself but couldn't get it to work, so wondering if you had any thoughts.

I.e. make a separate goroutine _each_ time we capture a frame or write a file.

muesli:
> On a semi-related note, would we be able to make the screenshot capturing stuff faster using goroutines? I tried playing around with it a bit myself but couldn't get it to work, so wondering if you had any thoughts.

Might make the timing easier, but also complicates situations where you can't keep up with the requested interval. I'll take a closer look!

maaslalani:
> > On a semi-related note, would we be able to make the screenshot capturing stuff faster using goroutines? I tried playing around with it a bit myself but couldn't get it to work, so wondering if you had any thoughts.
> 
> Might make the timing easier, but also complicates situations where you can't keep up with the requested interval. I'll take a closer look!

I believe this is similar to what I was suggesting: https://github.com/charmbracelet/vhs/pull/110

PR #100, opened 2022-10-30
fix: don't print errors twice

muesli:
We handle and print errors in main. Disable cobra's own error printing so they don't show up twice.


caarlos0 (approved):
was doing the same thing in another PR :D

muesli:
> was doing the same thing in another PR :D

oops, sorry, didn't see that!

maaslalani:
Thank you!

PR #102, opened 2022-10-30
feat: allow alternative outputs for serve

### LINKED ISSUE #59: Allow for multiple files/types in vhs serve
Currently `vhs serve` forces and returns a gif regardless of `Output`. 
It would be interesting to allow for multiple (or at least choose an alternative), which would then be streamed back as a single archive of files.

From a glance at the code I think it should be possible to check if there are multiple outputs and, if so, create an archive to return instead of the bytes of a single file.

-----

I would be interested in getting my feet a little wet if you find this proposal worth pursuing. :slightly_smiling_face:

jolheiser:
Partially resolves #59

I still think it would be nice to allow sending back an archive, but for now this will suffice.

One question is how do we define priority? e.g. a tape file has outputs for multiple formats, which should be chosen?

Currently based on this PR, it would be `mp4` -> `webm` -> `gif` (default)

### DIFF
--- a/serve.go
+++ b/serve.go
@@ -90,13 +90,24 @@ var serveCmd = &cobra.Command{
 
 						//nolint:gosec
 						rand := rand.Int63n(maxNumber)
-						tempFile := filepath.Join(os.TempDir(), fmt.Sprintf("vhs-%d.gif", rand))
+						tempFile := filepath.Join(os.TempDir(), fmt.Sprintf("vhs-%d", rand))
 						defer func() { _ = os.Remove(tempFile) }()
 						errs := Evaluate(s.Context(), b.String(), s.Stderr(), func(v *VHS) {
-							v.Options.Video.Output.GIF = tempFile
-							// Disable generating MP4 & WebM.
-							v.Options.Video.Output.MP4 = ""
-							v.Options.Video.Output.WebM = ""
+							var gif, mp4, webm string
+							switch {
+							case v.Options.Video.Output.MP4 != "":
+								tempFile += ".mp4"
+								mp4 = tempFile
+							case v.Options.Video.Output.WebM != "":
+								tempFile += ".webm"
+								webm = tempFile
+							default:
+								tempFile += ".gif"
+								gif = tempFile
+							}
+							v.Options.Video.Output.GIF = gif
+							v.Options.Video.Output.MP4 = mp4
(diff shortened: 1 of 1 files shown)


serve.go
```diff
@@ -92,13 +92,25 @@ var serveCmd = &cobra.Command{
...
+								tempFile += ".webm"
+								webm = tempFile
+							default:
+								tempFile += ".gif"
+								gif = tempFile
+
+							}
+							v.Options.Video.Output.GIF = gif
+							v.Options.Video.Output.MP4 = mp4
+							v.Options.Video.Output.WebM = webm
```
maaslalani, line 113:
```suggestion
							switch {
							case v.Options.Video.Output.MP4 != "":
							    v.Options.Video.Output.MP4 = tempFile + ".mp4"
							case v.Options.Video.Output.WebM != "":
							    v.Options.Video.Output.WebM = tempFile + ".webm"
							default:
							    v.Options.Video.Output.WebM = tempFile + ".gif"

							}
```

serve.go
```diff
@@ -90,13 +90,25 @@ var serveCmd = &cobra.Command{
...
+							case v.Options.Video.Output.MP4 != "":
+								tempFile += ".mp4"
+								mp4 = tempFile
+							case v.Options.Video.Output.WebM != "":
+								tempFile += ".webm"
+								webm = tempFile
+							default:
+								tempFile += ".gif"
+								gif = tempFile
+
```
maaslalani, line 107:
```suggestion
							case v.Options.Video.Output.MP4 != "":
								mp4 = tempFile + ".mp4"
							case v.Options.Video.Output.WebM != "":
								webm = tempFile + ".webm"
							default:
								gif = tempFile + ".gif"
```

jolheiser: I would need to try this locally, but iirc I tried this initially and `ffmpeg` complained (at least about gifs) because the output file had an incorrect extension.

jolheiser: Ah wait, I need to revert this. 
`tempFile` is still used when cleaning up the temporary file

maaslalani:
Is this one ready for testing @jolheiser?

jolheiser:
Should be, yes. I opted to force push because I had some conflicts and LFS didn't want to play nice on Windows 😅

maaslalani (approved):
Works great! Thank you @jolheiser!

PR #108, opened 2022-10-30
fixes #62: allow numeric filenames

### LINKED ISSUE #62: Allow numeric filenames as output (`1.gif`)
Using `Output 1.gif` in a .tape file breaks with a friendly error.

```
File: test.tape
  1 │ Output 1.gif
      ^^^^^^ Expected file path after output

  1 │ Output 1.gif
             ^^ Invalid command: 1.

  1 │ Output 1.gif
               ^^^ Invalid command: gif

Error: parse error
```

Perhaps it could be allowed?
(Of course, adding an alphanumeric character to the path works.)

abhijit-hota:
fixes #62

### DIFF
--- a/parser.go
+++ b/parser.go
@@ -156,9 +156,12 @@ func (p *Parser) parseKeypress(ct TokenType) Command {
 func (p *Parser) parseOutput() Command {
 	cmd := Command{Type: OUTPUT}
 
+	// If the name starts with a digit or any other non string character,
+	// add that to the next token for the complete file name/path
 	if p.peek.Type != STRING {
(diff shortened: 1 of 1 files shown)


maaslalani:
Hey @abhijit-hota, thank you for the PR. I think though this is not the correct approach for solving the numeric file type prefix, It's currently possible to do:

```
Output "1.gif"
Output ./1.gif
```

That being said I will close this PR for now since it technically is possible to have numeric file names (just need to be in quotes, which I personally think is reasonable).

I may try and fix the parser to allow bare numeric names in the future though.

PR #112, opened 2022-10-31
fix: use cmd as default on windows

### LINKED ISSUE #93: Windows: Bash just throw back if WSL not installed
Hey! Nice tool.
But I found one painful bug on Windows.
This bug is not entirely obvious, because the build is successful:
```
File: demo.tape
Output .gif demo.gif
Set FontSize 46
Set Width 1200
Set Height 600
Type echo 'Welcome to VHS!'
Sleep 500ms
Enter 1
Sleep 5s
Creating GIF...
```

But at the output gif we see this:
demo

I found that this is due to the fact that WSL is not installed on my machine, and `bash` command just throws me back out:
bash

The solution, I think, is obvious to you: run `cmd` or `powershell` instead of `bash` on Windows machines.
https://github.com/charmbracelet/vhs/blob/cf3919b5f5f6f78ff39c08aa28fd4a2a8b071f87/tty.go#L34

I know that you are preparing a feature (Set-Shell) but keep in mind that if the Set-Shell was not explicitly specified in the file, it should be selected automatically depending on the OS.

Thank you!
Have a nice day!

caarlos0:
fixes #93

can someone on windows test?

### DIFF
--- a/tty.go
+++ b/tty.go
@@ -13,6 +13,7 @@ import (
 	"fmt"
 	"net"
 	"os/exec"
+	"runtime"
 )
 
 // randomPort returns a random port number that is not in use.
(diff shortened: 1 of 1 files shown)


tty.go
```diff
@@ -24,14 +25,20 @@ func randomPort() int {
...
 		"-t", "cursorBlink=true",
 		"-t", "customGlyphs=true",
-		"bash", "--login",
-	)
+	}
+	switch runtime.GOOS {
+	case "windows":
+		args = append(args, "cmd.exe")
+	default:
+		args = append(args, "bash", "--login")
```
caarlos0, line 39: after #55 we can probably default to just `sh`

maaslalani:
Thanks, Carlos! Looks perfect.

tty.go
maaslalani: Sounds like a plan!

maaslalani (approved):

Delta456:
> fixes #93
> 
> 
> can someone on windows test?

I will test this later...

Rebzzel:
> fixes #93
> 
> can someone on windows test?

It works, but it looks strange:
cmd

I think better to use `powershell` instead:
powershell

PR #113, opened 2022-10-31
feat: Set LoopOffset

### LINKED ISSUE #8: Support rendering offset
Support starting the rendering a few seconds into the recording, then looping around from the end to the beginning.

The idea behind this is that it generates a more reasonable thumbnail when sharing the gif on various platforms. The viewer won't typically notice this *if* the output is running as a continuous loop.

griimick:
Initial implementation after discussion https://github.com/charmbracelet/vhs/issues/8#issuecomment-1296319441

Closes #8


video.go
```diff
@@ -37,6 +37,24 @@ type VideoOutputs struct {
 	MP4  string
 }
 
+// LoopOffsetOptions is a set of options for rendering loop offset
+type LoopOffsetOptions struct {
+	frames     int
```
maaslalani, line 42: I think we can actually remove the # of frames options for a few reasons:

1. It's nice to have only one way of doing things (percentage loop offsets) it will make documentation and code nicer.
2. I doubt anyone will need to get so precise as to specify the frame count (percentages are better) since they won't know the frame count or frame number in advanced anyway.
3. Even if people did know the exact frame number they wanted VHS isn't deterministic so these frame counts can change based on the machine running it so it would be more repeatable since percentages would work across framerates (and work well with the `Set Framerate` setting.

griimick: I agree. I myself ended up using percentages for testing. 

Does using `VHS.Options.LoopOffset` to store percentage sound good? I will get rid of `LoopOffsetOptions` struct altogether.

maaslalani: Yep, that's perfect!

griimick: Should we keep `%` token for semantics? 
```elixir
Set LoopOffset 20.45 # implicit 
# OR
Set LoopOffset 20.45%
```

maaslalani:
Btw @griimick, we'll also need an update to the [tree-sitter-vhs](https://github.com/charmbracelet/tree-sitter-vhs/blob/main/grammar.js) grammar.js. You're welcome to do that if you want, or we can update that ourselves once this is merged!

video.go
griimick: I went with keeping it optional.

maaslalani:
I saw that VSCode (VHSCode? :P) syntax yesterday. Really nice work! Thank you!

maaslalani:
@griimick Random thought, I wonder if symlinking would also be a good option if it works (that way you can still maintain the original frames) and you know where the loop offset starts (since the frames after the last frames are all symlinks to the previous frames?) What do you think?

video.go
```diff
@@ -58,6 +59,9 @@ const defaultPadding = 72
 const defaultPlaybackSpeed = 1.0
 const defaultWidth = 1200
 
+// DefaultStartingFrame is the index of the file to start the image sequence loop from
+const DefaultStartingFrame = 1
```
maaslalani, line 63: Can we make this unexported similar to the rest?

griimick:
> @griimick Random thought, I wonder if symlinking would also be a good option if it works (that way you can still maintain the original frames) and you know where the loop offset starts (since the frames after the last frames are all symlinks to the previous frames?) What do you think?

I like the idea and also confirm it works (push commit?). The original frame sequence is preserved, especially for cases where a user is generating a `.png` frame sequence as output. 

But I am unsure which order the user really expects when applying offset to the loop.

video.go
griimick: I think I can. Let me push a commit to address this.

maaslalani:
@griimick Sounds good, we can leave the symlinks for now. I think the simpler and more compatible, the better. Are you able to rebase this branch off master to get it ready for merging?

maaslalani:
Looking good @griimick, just played around with it and it's fantastic. Will push up some additional cleanup and documentation

muesli:
Thank you! <3

griimick:
Just went through all the follow-up changes. Thanks, everyone.

melMass:
Would be nice to document it in the list of commands, quite useful thanks!

PR #116, opened 2022-10-31
Add support for variables, loops and functions with go template

metafates:
Hi charm! I've been using your tool lately and it's awesome! 
One thing I was missing is more programmatic approach.

This PR executes [text/template](https://pkg.go.dev/text/template) before parsing the tape further. That way it's possible to define variables, use loops and anything else go templates offer. I also added 2 basic functions for template's FuncMap such as `repeat` to repeat string N times and `add` to sum numbers. But of course more functions could be added.

Usage Example
```elixir
# We can define 'strings' function in template's FuncMap to create string arrays
{{ $required := strings "echo" "gum" "gh" }}

{{ range $bin := $required  }}
Require {{  $bin  }}
{{ end }}
```

Let me know what you think! 🐳

### DIFF
--- a/evaluator.go
+++ b/evaluator.go
@@ -16,6 +16,11 @@ type EvaluatorOption func(*VHS)
 // Evaluate takes as input a tape string, an output writer, and an output file
 // and evaluates all the commands within the tape string and produces a GIF.
 func Evaluate(tape string, out io.Writer, opts ...EvaluatorOption) error {
+	tape, err := ExecuteTemplate(tape)
+	if err != nil {
+		return err
+	}
+
 	l := NewLexer(tape)
 	p := NewParser(l)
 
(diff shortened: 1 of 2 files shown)


maaslalani:
Hey @metafates, this is super clever! I'm quite impressed, very efficient solution to the problem of having variables and loops, etc... Nice work!

Fortunately, we do have plans of introducing Loops and Variables to the VHS / `.tape` language so you can look forward to that! That being said, I'll close this PR but you should for sure continue using Go templates (or any other templating engine for that matter) with tape files until we've implemented it 😄 

Thanks for the PR, we really appreciate it ❤️

caarlos0:
also worth nothing: `envsubst`, `jinja` et al can still be used and pipe through, e.g.:

```sh
cat file.tap | some_template_engine | vhs -
```

PR #117, opened 2022-10-31
feat: use powershell as default shell on windows>=10 + compile-time optimizations

Rebzzel:
Using `cmd` instead of `powershell` leads to a strange render:
cmd

I also see no reason to set default shell in runtime.
This pull request fixes this, and leaves the option to use the tool on older versions.

### DIFF
--- a/tty_windows.go
+++ b/tty_windows.go
@@ -1,0 +1,19 @@
+//go:build windows
+// +build windows
+
+package main
+
+import "golang.org/x/sys/windows"
+
+func defaultShellWithArgs() []string {
+	shell := "cmd"
+	major, _, _ := windows.RtlGetNtVersionNumbers()
+
+	if major >= 10 {
+		shell = "powershell"
+	}
+
+	return []string{
+		shell,
+	}
+}
--- a/tty_windows.go
+++ b/tty_windows.go
@@ -6,14 +6,10 @@ package main
 import "golang.org/x/sys/windows"
 
 func defaultShellWithArgs() []string {
-	shell := "cmd"
 	major, _, _ := windows.RtlGetNtVersionNumbers()
-
 	if major >= 10 {
-		shell = "powershell"
+		return []string{"powershell"}
 	}
 
-	return []string{
-		shell,
-	}
+	return []string{"cmd"}
 }
--- a/tty.go
+++ b/tty.go
@@ -13,7 +13,6 @@ import (
 	"fmt"
 	"net"
 	"os/exec"
-	"runtime"
 )
 
 // randomPort returns a random port number that is not in use.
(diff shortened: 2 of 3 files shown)


tty_windows.go
```diff
@@ -0,0 +1,25 @@
...
+package main
+
+import (
+	"strings"
+	"strconv"
+	"github.com/shirou/gopsutil/host"
+)
+
+func defaultShellWithArgs() []string {
+	hostInfo, _ := host.Info()
```
Delta456, line 13: I don't think you actually need `github.com/shirou/gopsutil/host` module because the same can be done with `golang.org/x/sys/windows`'s `RtlGetNtVersionNumbers` function.

caarlos0 (approved):
thanks, this looks good

tty_windows.go
muesli: I tend to agree, we can probably do this without extra deps.

caarlos0:
refs #112
refs #93

caarlos0:
PS: that weird formatting error you say will still happen in this, at least until #55 is merged

Rebzzel:
> PS: that weird formatting error you say will still happen in this, at least until #55 is merged

Do you mean runtime path in output?

Rebzzel:
@Delta456's review applied, now pull request has no extra deps

griimick:
> Do you mean runtime path in output?

default `cmd.exe` greeting, command to modify prompt and clear screen.

Rebzzel:
> default `cmd.exe` greeting, command to modify prompt and clear screen.

Yes, that's why `powershell` is used here by default on the OS where it is available.

demo

https://github.com/charmbracelet/vhs/blob/bf3b72d5f0edf722a411dfb06d1f24ef70be096a/tty_windows.go#L10-L13

caarlos0 (approved):

Rebzzel:
The last commit may seem strange, but for some reason linter stopped at `tty_windows.go`. 
I fixed it with `gofmt -w tty_windows.go`, but now my file didn't see `git` 👀 .
In general, I had to add some empty lines to make this commit 🌚

Rebzzel:
Hey, guys, who doesn't know what could have happened?
At the moment, `golangci-lint` gives me an error in almost all files:
```
command.go
command_test.go
embed.go
error.go
evaluator.go
keys.go
lexer.go
lexer_test.go
main.go
man.go
parser.go
parser_test.go
serve.go
serve_unix.go
serve_windows.go
style.go
syntax.go
testing.go
theme.go
token.go
tty.go
tty_unix.go
vhs.go
video.go
```

After `gofmt -r '(a) -> a' -l *.go`, the problem is fixed but `git` sees changes only changes in two files:
```
lexer.go
tty_unix.go
```

Also, `git diff` outputs an empty string to them:

```
$ git diff lexer.go tty_unix.go

```

I can understand that the files I created could have been saved in some wrong format, but why did it affect the rest of the files? :\

Rebzzel:
I have no idea what it was, so I had to roll back to fix it.

PR #118, opened 2022-10-31
docs(readme): add MacPorts install info

herbygillot:
On macOS, `vhs` can now be installed via MacPorts: https://ports.macports.org/port/vhs/


maaslalani:
Thank you so much @herbygillot!

PR #120, opened 2022-10-31
Add example for onefetch

o2sh:
I was trying to use the `vhs.ascii` logo but it fails to execute the command.

Expected output (on iterm):

Screenshot 2022-11-01 at 00 21 38

With vhs:

Screenshot 2022-11-01 at 00 23 43

Amazing tool BTW 👍


maaslalani:
Hey! Thanks so much for this PR. One fetch looks like an awesome tool. We already have a very similar example with neofetch so it's nice to avoid duplicate / very similar examples. I will close this for now but we can revisit if  the example shows different functionality of VHS.

Thanks again for this PR ❤️

PR #124, opened 2022-11-01
feat: allow external context to be passed to Evaluate

muesli:
This makes vhs jobs cancelable and removes the need for a separate signal handler in the serve command.

It is a slight behavioral change however: we now actively cancel ongoing connections when the sever gets asked to terminate. I intend to address this in a separate PR which sets a timeout for incoming connections.


maaslalani (approved):

maaslalani:
Looks great, feel free to merge.

PR #125, opened 2022-11-01
fix: split up video rendering & cleanup

muesli:
This ensures canceled jobs will always clean up temporary files on teardown.


maaslalani (commented):
Thank you!

PR #126, opened 2022-11-01
chore: simplify vhs.Record

muesli:
Avoids nested ifs, simplifies interval & timeout handling.


maaslalani (approved):
@muesli just to confirm this only sleeps for the interval - the time elapsed for the frame capture right?

From what I understand, that is the case but just want to confirm.

muesli:
@maas Correct, every time we try to capture a frame, we'll record the last start time. On the next iteration of the loop, we'll wait `interval` - `time elapsed since last start`. This essentially replaces the need for a manual `time.Sleep` and lets `Record` concurrently wait for the next interval or the context being cancelled.

PR #129, opened 2022-11-02
fix: never os.Exit, return errors

muesli:
This moves the error handling out of the Evaluate method: the callers are responsible for handling (printing) the errors.

This removes a bit of duplicate code and prevents the server from being terminated on encountering the first parser error.


maaslalani (approved):

serve.go
```diff
@@ -91,17 +91,16 @@ var serveCmd = &cobra.Command{
 						//nolint:gosec
 						rand := rand.Int63n(maxNumber)
 						tempFile := filepath.Join(os.TempDir(), fmt.Sprintf("vhs-%d.gif", rand))
-
-						// Clean up the temp file eventually.
 						defer func() { _ = os.Remove(tempFile) }()
-
-						err = Evaluate(s.Context(), b.String(), s.Stderr(), func(v *VHS) {
+            
+						errs := Evaluate(cmd.Context(), b.String(), s.Stderr(), func(v *VHS) {
```
maaslalani, line 96:
```suggestion
						errs := Evaluate(s.Context(), b.String(), s.Stderr(), func(v *VHS) {
```

serve.go
```diff
@@ -91,17 +91,15 @@ var serveCmd = &cobra.Command{
...
-						err = Evaluate(s.Context(), b.String(), s.Stderr(), func(v *VHS) {
+						errs := Evaluate(s.Context(), b.String(), s.Stderr(), func(v *VHS) {
 							v.Options.Video.Output.GIF = tempFile
 							// Disable generating MP4 & WebM.
 							v.Options.Video.Output.MP4 = ""
 							v.Options.Video.Output.WebM = ""
 						})
-						if err != nil {
+						printErrors(s.Stderr(), b.String(), errs)
+						if len(errs) > 0 {
```
maaslalani, line 102:
```suggestion
						
						if len(errs) > 0 {
						  printErrors(s.Stderr(), b.String(), errs)
```

PR #131, opened 2022-11-02
fix: never os.Exit, return error

muesli:
This is a bit of an edge case, as the themes.json file is embedded and technically should never fail to parse, but even then it's probably a bad idea to tear down the entire process without any deferred cleanups.

### DIFF
--- a/command.go
+++ b/command.go
@@ -333,19 +333,8 @@ func getTheme(s string) (Theme, error) {
 	case '{':
 		return getJSONTheme(s)
 	default:
-		return getNamedTheme(s)
+		return findTheme(s)
 	}
-}
-
-func getNamedTheme(s string) (Theme, error) {
-	theme, suggestions, ok := findTheme(s)
-	if !ok && len(suggestions) > 0 {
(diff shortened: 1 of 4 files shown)


command.go
```diff
@@ -338,7 +338,10 @@ func getTheme(s string) (Theme, error) {
 }
 
 func getNamedTheme(s string) (Theme, error) {
-	theme, suggestions, ok := findTheme(s)
+	theme, suggestions, ok, err := findTheme(s)
```
maaslalani, line 341: Do we need both `ok` and `err`? I assume `err != nil` is equivalent to `ok`?

muesli: There's a difference, even though we could unify the `bool` and `err` by using a typed error:

The `err` indicates a fatal error, e.g. when `themes.json` is corrupted. But there's also the situation where we just can't find the requested theme: in that case the `bool` is false, and we return suggestions.

maaslalani: Gotcha, yeah that sounds good then.

maaslalani (approved):
Amazing 🙌, Thanks @muesli!

caarlos0 (approved):

PR #135, opened 2022-11-02
VHS Record Feature (`vhs record`)

maaslalani:
This PR allows users to automatically write `.tape` files by recording their terminal.

To use simply:

```
vhs record > myrecording.tape
```

Then, use the terminal as you wish. VHS will record commands, typing, and
sleeps and write them to a tape file.

Once the tape file is written, you may want to tweak it manually before
running:

```
vhs < myrecording.tape
```

### DIFF
--- a/examples/neofetch/colorize-ascii.go
+++ b/examples/neofetch/colorize-ascii.go
@@ -1,4 +1,4 @@
-package main
+package neofetch
 
 import (
 	_ "embed"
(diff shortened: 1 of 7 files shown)


paralin:
Great! Thanks

paralin:
Posting this here as well: https://github.com/rsteube/cast2tape

PR #140, opened 2022-11-03
fix(examples): output file name

aymanbagabas:
bash examples overwrites zsh


maaslalani:
Good catch! Thank you

PR #146, opened 2022-11-05
Cross-platform fixes + new features

### LINKED ISSUE #39: Strange prompt output at the start of the gif
I attempted to put together a gif to demo my crate, [woodo](https://lib.rs/woodo): https://github.com/knpwrs/woodo

Here is my script:

```tape
Output woodo.gif

Set FontSize 32
Set Width 800
Set Height 1200

Type woodo
Sleep 500ms
Enter
Sleep 500ms
Type sudo woodo
Sleep 500ms
Enter
Type lolnope
Enter
Sleep 5s
```

Here is the output gif:

woodo

The gif starts out with 11 frames, ending with:

image

Before transitioning to:

image

Rebzzel:
Hey guys, I have a lot of things for you.
To begin with, the tool has big problems with cross-platform and not only, so I suggest you look at my solutions to these problems.

## Entry Points Introduction
Allows to select the environment in which subsequent commands will work.
Something like a replacement for 6466fa95686a2e3c245bfbe8245741221660ba63 but with more features.

## "Lazy Shells" Introduction
_From the point of view of the language, these are just functions that return `Shell` structure._

This thing allows you to handle additional conditions, which for example helps to solve compatibility problems or detect if the user does not have a specific shell.

Example:
https://github.com/charmbracelet/vhs/blob/3797c980d7c84875fd34193b5c44865c3d9d022f/shell_windows.go#L16-L29

_Here we determine which version of PowerShell is available and, depending on this, we call the necessary parameters._

### Remove useless work (due to `feat: Set Shell` and features above)
See f525522172e0d9590943f0247ba4bcd2553b6a26.
Previously, it was necessary to solve the problem of working on Windows, as well as fix the output.
And now with the new features, it's just useless.

### Splitting the shell list into "Windows compatible" and "Unix compatible".
See [shell_windows.go](https://github.com/charmbracelet/vhs/blob/3797c980d7c84875fd34193b5c44865c3d9d022f/shell_windows.go) and [shell_unix.go](https://github.com/charmbracelet/vhs/blob/3797c980d7c84875fd34193b5c44865c3d9d022f/shell_unix.go)

### Fix incorrect powershell output (72d81d6b52a375d45dda97c26625f7b8335bd220)

Tape:
```elixir
Output demo.gif
Set Shell pwsh

Set FontSize 46
Set Width 1200
Set Height 600

# Type a command in the terminal.
Type "echo 'Welcome to VHS!'"

# Pause for dramatic effect...
Sleep 500ms

# Run the command by pressing enter.
Enter

# Admire the output for a bit.
Sleep 5s
```

Before:
frame223

After:
frame221

### Fix #39 (e8d5abf3beb889782ad3d66e810b9f0a75b7f3a8)
This problem arises due to the fact that we start recording before our "prepare" commands are executed. The next solution is more of a crutch, which in some cases may not help, but still better than nothing.

https://github.com/charmbracelet/vhs/blob/e8d5abf3beb889782ad3d66e810b9f0a75b7f3a8/vhs.go#L165-L167

First frame before:
frame1

First frame after:
frame1

###  Remove unlogical `pwsh != powershell` from b0c663c9bc8065d73412a442d8672f4fe59cffa0
Due to "Lazy Shells" now it's possible.

Before:
```elixir
# What to use?
Set Shell pwsh
# or
Set Shell powershell

# What a difference? Idk...
# HELP!
```

After:
```elixir
# 😎
Set Shell pwsh
```

### ~Another crutch~ Correction of possible errors related to the fact that logging into the shell may be too long
https://github.com/charmbracelet/vhs/blob/3797c980d7c84875fd34193b5c44865c3d9d022f/vhs.go#L132-L142

_At first it was one of the implementations of new features, but then I realized that it solves more things._

---

In the end, I want to say that I'm new to golang, so there may not be completely rational things here, so if anything, leave your reviews so that I can improve and learn new things)

Thank you!
Have a nice day!

### DIFF
--- a/command.go
+++ b/command.go
@@ -254,7 +254,7 @@ func ExecuteSetWidth(c Command, v *VHS) {
 // ExecuteSetShell applies the shell on the vhs.
 func ExecuteSetShell(c Command, v *VHS) {
 	if s, ok := Shells[c.Args]; ok {
-		v.Options.Shell = s
+		v.Options.Shell = s()
 	} else {
 		v.Options.Shell.Prompt = ""
 		v.Options.Shell.Command = c.Args
--- a/shell.go
+++ b/shell.go
@@ -2,44 +2,18 @@ package main
 
 // Supported shells of VHS
 const (
-	bash       = "bash"
-	cmdexe     = "cmd"
-	fish       = "fish"
-	powershell = "powershell"
-	pwsh       = "pwsh"
-	zsh        = "zsh"
+	bash   = "bash"
+	cmdexe = "cmd"
+	fish   = "fish"
+	pwsh   = "pwsh"
+	zsh    = "zsh"
 )
 
 // Shell is a type that contains a prompt and the command to set up the shell.
 type Shell struct {
-	Prompt  string
-	Command string
+	EntryPoint string
+	Prompt     string
+	Command    string
 }
 
-// Shells contains a mapping from shell names to their Shell struct.
-var Shells = map[string]Shell{
-	bash: {
-		Prompt:  "\\[\\e[38;2;90;86;224m\\]> \\[\\e[0m\\]",
-		Command: ` set +o history; unset PROMPT_COMMAND; export PS1="%s"; clear;`,
-	},
-	zsh: {
-		Prompt:  `%F{#5B56E0}> %F{reset_color}`,
-		Command: ` clear; zsh --login --histnostore; unsetopt PROMPT_SP; unset PROMPT; export PS1="%s"; clear`,
-	},
-	fish: {
-		Prompt:  `function fish_prompt; echo -e "$(set_color 5B56E0)> $(set_color normal)"; end`,
-		Command: `clear; fish --login --private -C 'function fish_greeting; end' -C '%s'`,
-	},
-	powershell: {
-		Prompt:  "Function prompt {Write-Host \"> \" -ForegroundColor Blue -NoNewLine; return \"`0\" }",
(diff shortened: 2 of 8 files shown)


muesli:
I feel like we're mixing together a couple of changes in a single PR here. Might be easier to review if we stick to one issue per PR.

griimick:
`pwsh` is used by newer cross-platform PowerShell 6+ whereas `powershell` is used for legacy one which comes installed by default in Windows 10 and above. They can coexist hence the difference in names.

Rebzzel:
> I feel like we're mixing together a couple of changes in a single PR here. Might be easier to review if we stick to one issue per PR.

I agree, but I think it's already too late...

Rebzzel:
> is used by newer cross-platform PowerShell 6+ (Core) which needs separate installation whereas `powershell` is used for legacy one which comes installed by default in Windows ~10 and above~ even older. They can coexist hence the difference in names.

Thank you, but I didn't ask, I was just simulating an ordinary developer who doesn't know the details. That's why it's a PR section, not an issue.

caarlos0:
I think most of this have been fixed on main, and there's a couple more improvements in #218... 

I like the idea of the magic text, although I believe #218 will make it unnecessary...

maaslalani:
Like Carlos mentioned, most of the fixes are currently on main, with the addition of starting shells with ttyd, I believe this PR can be closed. Let me know if there are any missing features that we still need to add here.

PR #147, opened 2022-11-06
GitHub syntax highlighting for .tape files

metafates:
Github's highlighting is powered by [github/linguist](https://github.com/github/linguist/blob/master/CONTRIBUTING.md#adding-a-language), and they allow to submit new languages if they're popular enough (used in >200 repos).

Until GItHub supports it, throwing this in a repo's .gitattributes will highlight it by pretending it's Elixir (as you do in the README):

```gitattributes
*.tape linguist-language=Elixir
```

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -4,3 +4,4 @@
 **/*.png filter=lfs diff=lfs merge=lfs -text
 themes*.json linguist-generated
 THEMES.md linguist-generated
+*.tape linguist-language=Elixir


maaslalani:
Hey @metafates, thank you so much for the PR! The problem with highlighting everything as elixir is that the highlighting breaks in some cases and looks quite bad. All the examples in the README were written in a way to not break the elixir highlighting but the examples in the examples folder were not.

Hopefully tape files will be found in lots of repos so that we can get highlighting via our grammar on `linguist`!

Issue #160, opened 2022-11-14
Docker image uses an alpine base and is incompatible with gum, other charm tools


taigrr:
The instructions show how to use a docker run command to use vhs ephemerally, but the image is based on alpine, which doesn't have the right libc / shared libraries.

I would recommend using a debian-based image instead, which should at least give better error messages (will warn if GLIBC is out of date rather than just "No such file or directory").

Ubuntu images will try to use a snapd version of chromium.
Debian does not have many of the fonts required.
I am submitting an example PR which uses alpine as a builder image to fetch the fonts, then passes them to a debian image so that we have glibc.

PR incoming.

PR #161, opened 2022-11-14
use debian for docker instead of alpine

taigrr:
See #160

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,11 +1,5 @@
 FROM tsl0922/ttyd:alpine as ttyd
-FROM alpine:latest
-
-# Create volume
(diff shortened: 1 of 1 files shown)


taigrr:
@maaslalani this is in reference to the conversation over on the discord help channel

willis81808:
using an image with this change solved a puzzling `No such file or directory` issue I was experiencing with VHS when generating a GIF from a shell script leveraging Gum

CaelFrost:
https://github.com/sgerrand/alpine-pkg-glibc
alternatively, if you wish to stick to alpine, would this work?

aymanbagabas (approved):

PR #178, opened 2022-11-25
feat: allow --output to override output location

maaslalani:
Allow `Output` location to be overridden by CLI flags:

Specify multiple outputs in the same flag.

```bash
vhs demo.tape --output out.gif,out.webm
```
  
Pass multiple flags (with shorthand):
  
```bash
vhs demo.tape -o out.gif -o out.webm
```

Note: this will override any `Output` commands set in the tape file.


maaslalani:
cc @mxcl

PR #182, opened 2022-11-29
Message about publishing GIFs

maaslalani:
Publish GIFs tip (when `VHS_PUBLISH`) is not set.

### DIFF
--- a/publish.go
+++ b/publish.go
@@ -10,6 +10,7 @@ import (
 	"path/filepath"
 
 	"github.com/charmbracelet/keygen"
+	"github.com/mattn/go-isatty"
 	gap "github.com/muesli/go-app-paths"
 	"github.com/spf13/cobra"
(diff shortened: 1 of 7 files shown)


publish.go
```diff
@@ -115,7 +116,9 @@ func sshSession() (*ssh.Session, error) {
 
 // Publish publishes the given GIF file to the web.
 func Publish(ctx context.Context, path string) (string, error) {
-	fmt.Println("Publishing GIF...")
+	if isatty.IsTerminal(os.Stdout.Fd()) {
+		fmt.Println("Publishing GIF...")
+	}
```
maaslalani, line 121: Added this so that people can simply do:

```
vhs publish demo.gif | pbcopy
```

and it will only copy the URL.

aymanbagabas (approved):

PR #183, opened 2022-11-29
fix(record): use backticks if a string contains double and single quotes

### LINKED ISSUE #181: Record mode cannot deal with quotes
What better way to report an issue than with a tape 🚀

```
$ vhs record > /tmp/test.tape
> echo '"'
> exit

$ cat /tmp/test.tape
Sleep 500ms
Type 'echo '"''
Enter

$ vhs < /tmp/test.tape
Sleep 500ms
Type echo  ''
Enter 1
Creating GIF...
```

Side note: quotes in general has been a bit of a pain point for me working with VHS.

maaslalani:
Fixes #181

### DIFF
--- a/record.go
+++ b/record.go
@@ -163,6 +163,9 @@ func inputToTape(input string) string {
 
 // quote wraps a string in (single or double) quotes
 func quote(s string) string {
+	if strings.ContainsRune(s, '"') && strings.ContainsRune(s, '\'') {
+		return fmt.Sprintf("`%s`", s)
+	}
 	if strings.ContainsRune(s, '"') {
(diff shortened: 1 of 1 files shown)


Airblader:
@maaslalani Not to be that guy, but if the input contains all three characters, this is still broken, right?

maaslalani:
That is correct, it would be broken, in that (hopefully rare) case we can split the string into multiple tokens and perform different commands.

Is this something you've run into in the past?

Airblader:
Fortunately not yet. I can see how it'd be premature handling this, but wondering if we couldn't at least error out (which should hopefully be easy to do).

Issue #179, opened 2022-11-27
Elegant file path checks


rm-dr:
If you give vhs an output path that doesn't exist, it won't notice until it calls `ffmpeg`. This is annoying and unintuitive, especially when your tape file takes a while to run.

**To Reproduce:**
1. `cd` to a directory that doesn't contain a directory named `examples`
2. `vhs new broken.tape`
3. `vhs < broken.tape`

Issue #173, opened 2022-11-24
Progress bars & Overlays


rm-dr:
Generalizing #164, it would be great to have a way to add "overlays" to `vhs` recordings.

This would allow for...
 - Verbose key names (As in `#164`)
 - Progress bars (It's not always clear where a gif begins and ends, especially in a `README`)
 - Emulated window frames and backgrounds, like [carbon](https://github.com/carbon-app/carbon)

adriens:
I 🗳️ for _"Emulated window frames and backgrounds, l"_

maaslalani:
This is a fantastic idea! We definitely want to add window frames + backgrounds and it would be really cool to have key overlays for tutorials.

rm-dr:
This is pretty easy to pull off with a few extra `ffmpeg` filters, especially with static images:


https://user-images.githubusercontent.com/96270320/204033970-6ffc1f97-bb03-4f28-8cfa-6f3307d90f39.mp4

adriens:
This is like a carbon ...but :100: :muscle: :weight_lifting: :1st_place_medal:

rm-dr:
I'm almost ready to submit [a PR](https://github.com/rm-dr/vhs/tree/image-background) for static image frames.

A few questions before I do so:
1. I've added `Set ImageFrame <file>` and `Set FrameSize <number>` settings. Should I keep these as settings, or should I make them standalone keywords?
2. I'd like to update [`README.md`](https://github.com/charmbracelet/vhs/blob/main/README.md) with an example, but the gifs there are all linked to `stuff.charm.sh`. Wouldn't it be better to have those directly linked to the repo, so that contributors that can't access `stuff.charm.sh` (like me) can add demos too? Maybe that's another issue.

PR #187, opened 2022-12-08
Added Decorations

rm-dr:
# Preview:
image

everything

```elixir
# Top gif
Output out.gif
Set FontSize 20
Set Width 600
Set Height 300
Set Padding 20

Set MarginFill image.jpg
Set MarginSize 30

Type "This is demo text"
Sleep 2s
```

```elixir
# Bottom gif
Output out.gif
Set FontSize 15
Set Width 600
Set Height 300
Set Padding 20

Set MarginFill "#79B8C3"
Set MarginSize 40

Set WindowBar Colorful
Set WindowBarSize 40

Set CornerRadius 10

Type "I have a bar, a margin, and rounded corners."
Sleep 2s
```

Added margins, round corners, and window bars. Addresses most of #173.

# Features:
- Adds settings `MarginSize`, `MarginFill`, `WindowBar`, `WindowBarSize`, `CornerRadius`
- Dimensions of the output are exactly `width` and `height`, terminal size is calculated
- Errors are raised when `width` or `height` are too small

# Other notes:
 - This does not gracefully exit when a margin image doesn't exist. I left that out intentionally, should be fixed with #179.
- I did not add an example to the main `README.md` since the examples there are linked from `stuff.charm.sh`, and I do not have access to that. It might be a good idea to remove the examples in that file completely, since they're already in `examples/*`. There's no need to have *two* copies of every example tape!
- It may be good to add a warning about gif palettes to the docs. Margin images with many colors turn out pretty ugly.


adriens:
OMG :heavy_heart_exclamation: What a beauty :clap:

maaslalani:
This is awesome, @rm-dr! Thank you so much for this PR.

Some feedback on this feature and thinking about the language interface: the thing I think of when I see this is an image background, rather than a frame. When I think of a frame I moreso think of the border of the terminal (such as the green, red, yellow dots).

So, I think maybe we should rename this `Background`. I can see a future where we also also an option for terminal frames + backgrounds + margin. Something like:

```elixir
Set Frame MacOS
Set Background https://example.com/link-to-some-cool-background.png
Set Background #FFF # perhaps this can also take a color?
Set Margin 40 # the terminal frame will have 40px of background surrounding it
Set Padding 20 # the content will have 20px of the terminal background color surrounding it
```

adriens:
OH BOY, OH BOY ❣️ 🏋️ 🧑‍🎨

rm-dr:
Yes, that sounds like a good idea.

`ImageFrame` and `FrameSize` should be renamed, but `Background` is already used in multiple places to refer to the *terminal's* background (`Options.Video.BackgroundColor` and `Options.Theme.Background`, for example).

Adding an `Options.Video.Background` would make things a bit confusing.

We could use a different name internally, like `MarginBackground?`
That might be a clearer, or it might be more confusing than before.

rm-dr:
I renamed the setting keys and added static color support. It would be nice to add a format check for colors so a malformed hex string won't give a huge, hard-to-read ffmpeg error.

maaslalani:
This is really awesome @rm-dr, thank you _so_ much!

examples/settings/set-margin.tape
```diff
@@ -0,0 +1,13 @@
+Output "examples/settings/set-margin.gif"
+Set FontSize 20
+Set Width 600
+Set Height 300
+Set Padding 20
+
+# You can also use an image:
+# Set MarginFill "path/to/image.png"
+Set MarginFill "#79B8C3"
+Set MarginSize 20
```
maaslalani, line 10: I _think_ we'll want to call these `Background` and `Margin` externally. Internally, in the code, we can keep these names.

```suggestion
Set Background "#79B8C3"
Set Margin 20
```

video.go
```diff
@@ -78,6 +85,148 @@ func DefaultVideoOptions() VideoOptions {
 	}
 }
 
+// BuildFFopts builds an ffmpeg command from a VideoOptions
+func buildFFopts(opts VideoOptions, targetFile string) []string {
```
maaslalani, line 89: Nice, thanks for making this better!

maaslalani:
Just as a heads up, we will likely need this error to be a `ParserError` so that it does the nice underlining for the color token.

image


Such as like this:

image

Let me know if you need any pointers on how to do that (or if you want me to implement this myself)

rm-dr:
I just added parse-time color checking, and may have found a minor bug:

Error highlighting doesn't account for quoted strings!
image

examples/settings/set-margin.tape
rm-dr: Are you sure? I think that the `Background` setting key is a bit misleading, since backgrounds are usually "interior & padding" areas, not "margin" areas.

For example, this is how CSS works ([jsfiddle](https://jsfiddle.net/x3wcuabm/4/)).

Maybe run a poll on discord?

maaslalani: Yeah, actually I think you are right, background might suggest the terminal background.

Another approach we could go for is calling it Border, what do you think?

```
Border 20px "#FFF"
```

or 

```
BorderSize 20px
BorderFill "#FFF"
```

maaslalani:
> I just added parse-time color checking, and may have found a minor bug:
> 
> Error highlighting doesn't account for quotes surrounding strings! image

Thanks! That looks awesome.

Yeah the off by one error definitely looks like a bug, should be an easy solve. It's unfortunate that `#` is the comment character which means the color must be quoted :/

examples/settings/set-margin.tape
maaslalani: Perhaps `Margin` is the correct name, Border could be used in conjunction with the Margin I guess.

rm-dr: Yeah, I think "Margin" makes the most sense
Border is terminal frame, padding is padding, etc

Just like the CSS box model we're used to!

maaslalani:
@rm-dr The new rounded corners and terminal frames looks so awesome! Thank you so much!

winder:
I gave this branch a try today. It worked as expected and the rendered recording looks great. Despite that there were errors reported:
```
Couldn't draw Bar, something is very wrong.
Couldn't draw CornerMask, something is very wrong.
```

I used the window configuration you have above:
```
Set MarginFill "#79B8C3"
Set MarginSize 40

Set WindowBar Colorful
Set WindowBarSize 40

Set CornerRadius 10
```

rm-dr:
Stupid bug. I put a `==` where a `!=` should be. Give me a second...

rm-dr:
That's odd. Yesterday's changes could not possibly affect margin scaling (or anything else, in fact).

You're right, vhs should force aspect ratio if the margin image does not match the video. I tried to reproduce this today, but couldn't get your results. Can you send your image/config/system info?

My test files are below.



```elixir
Output test.gif
Set FontSize 20
Set Width 600
Set Height 300
Set Padding 20

Set MarginFill test.jpg
Set MarginSize 50

Type "I have a margin!"
Sleep 2s
```

test

rm-dr:
I know this is a big addition to `vhs`, but it would be nice to have a merge ETA.

Is there anything else that needs to be done?

maaslalani:
Hey @rm-dr, thanks again so much for this awesome PR, I'll try and give it a first pass review today and try out all the features!

maaslalani:
Amazing work! This has been merged in via: https://github.com/charmbracelet/vhs/pull/274

PR #199, opened 2022-12-21
fix: ensure directory is present before creating output

maaslalani:
Fixes poor error message when directory doesn't exist for output.


maaslalani:
Fixes #179

PR #212, opened 2023-01-13
fix: zsh, bash and fish: do not use user/system rcs

caarlos0:
assuming that probably in https://github.com/charmbracelet/vhs/issues/39 the problem is zsh is using the user/rc config, this should prevent that.

also did the same for bash and fish.


maaslalani (approved):
Thank you! Feel free to merge when green!

caarlos0:
also added a small fix to prevent the fish setup to be added to the history

PR #214, opened 2023-01-16
fix: start on sh instead of bash, fix pwsh

caarlos0:
- start on plain old `sh` and then fire the shell we want (follow up from #212)
- make fish example more similar to the others
- removed duplicated example
- fixed pwsh escape sequences
- do not use pwsh/powershell profile (follow up #212 too)
- pushed gif for pwsh example


maaslalani:
Awesome!

PR #218, opened 2023-01-30
feat: init ttyd with the user-set shell

caarlos0:
based on #214 and #212, for cases like https://github.com/charmbracelet/vhs/discussions/169

the idea is to parse the tape earlier on, evaluate the set shell, and start ttyd with the correct shell already.

also, moved some things around, for instance, instead of `setopt blah; clear`, add it to a rcfile and use it instead.

with this, ttyd will already start with the correct shell, already set up, and we shouldn't need to `clear` here and there, thus preventing the appearances of the setup step in the gif.

that said, the impl is still very rough, and cmdexe is still missing an impl.


maaslalani:
This makes sense to  me, I like this approach!

caarlos0:
this is a bigger endeavor than #214, but I think it'll work better in the end.

not sure if we merge #214 first, and then work on this with time, or give this one a bit more time and be over with it... thoughts?

maaslalani:
Yeah it's fine to sit on this one, I think #214 we can merge first!

caarlos0:
I just wasn't able to test on windows/cmd... not even ttyd works with cmd on my VM... I couldn't figure out why... if someone with windows can give it a go it would be amazing.


anyway, this should improve overall stability of VHS's output, as well as maybe speed it up a little bit.

Rebzzel:
> I just wasn't able to test on windows/cmd... not even ttyd works with cmd on my VM... I couldn't figure out why... if someone with windows can give it a go it would be amazing.

a
set-shell-cmd

caarlos0:
Oh cool, os it just didn't set the prompt 🤔

Will investigate more later 🙏

Thanks for giving it a try ❤️

caarlos0:
okay, found out why my ttyd wasn't working on windows: https://github.com/tsl0922/ttyd/issues/1064

also fixed the prompt on cmd, should be all good now!

thanks everyone!

PR #222, opened 2023-02-12
feat: add loglevel flag for silent output

### LINKED ISSUE #189: Make vhs be quiet
Exactly what it says on the tin.

Any self-respecting cli tool has a flag akin to `--quiet` that suppresses non-critical output. \
Since VHS is likely to be used in build scripts, it needs this feature more than most.

elmarsan:
Closes #189 

I've added `loglevel` flag it accepts `quiet` and `verbose` by default it's `verbose`.

In case you execute with quiet, the terminal will hide logs. E.g: `vhs demo.tape --loglevel=quiet `


publish.go
```diff
@@ -122,15 +122,15 @@ func sshSession() (*ssh.Session, error) {
 }
 
 func publishShareInstructions(url string) {
-	fmt.Println("\n" + GrayStyle.Render("  Share your GIF with Markdown:"))
```
bashbunni, line 125: Need to check if this breaks `vhs publish` as it depends on outputting the shareable URL

record.go
```diff
@@ -75,7 +76,7 @@ func Record(cmd *cobra.Command, args []string) error {
 	}
 
 	if err := pty.InheritSize(os.Stdin, terminal); err != nil {
-		log.Printf("error resizing pty: %s", err)
+		logger.Printf("error resizing pty: %s", err)
```
bashbunni, line 79: This is probably ok to be ignored for the `--quiet` flag, but would be helpful for the `--debug` mode

bashbunni (commented):
Nice work, left a couple notes for next reviewer to hopefully speed up the process. Thanks for the hard work :)

record.go
elmarsan: Yeah, I agree with this point.
Maybe we could have a better description about the logger modes and continue improving this point in future pr's.

bashbunni:
I'm thinking that the `vhs publish demo.gif --loglevel=quiet` should just output the sharable link without the other embeds... That might make it more versatile if people are hoping to use it for scripting and whatnot.

publish.go
bashbunni: The logger doesn't get initialized at the beginning of the `publish` subcommand, so just PR'd your branch to correct that :) Looks great!

bashbunni (changes requested):
Hey @elmarsan, 

I spoke with @maaslalani (the author) and here's some feedback on this feature:
- [ ] `-v` verbose. `-q` or `--quiet` quiet, get rid of "loglevel="
- [ ] have quiet only output the shareable link for `vhs publish`

Let us know if you're comfortable making those changes :smiley:

elmarsan:
Hey @bashbunni thanks for the feedback!

I'll do the changes asap.

elmarsan:
Hey @bashbunni 

I've applied some changes in order to achieve both points you mentioned above.
Could give it another look, please?

Thanks!

logger.go
```diff
@@ -0,0 +1,50 @@
+package main
+
+import (
+	"io"
+	"log"
+	"os"
+)
+
+// LogLevel represents custom type for configure
+type LogLevel int
```
caarlos0, line 10: its not really used as logger though, more likely general CLI output... 

with that in mind, maybe we could call this `output` or something?

then we merely have quiet output (just errors) or not (full output) 🤔

main.go
```diff
@@ -35,13 +35,25 @@ var (
 	publishFlag bool
 	outputs     *[]string
 
+	// Logging flags
+	verboseFlag bool
+	quietFlag   bool
```
caarlos0, line 40: probably we could just add the `--quiet` flag, as verbose is the default

main.go
```diff
@@ -35,13 +35,25 @@ var (
...
 		Args:          cobra.MaximumNArgs(1),
 		SilenceUsage:  true,
 		SilenceErrors: true, // we print our own errors
 		RunE: func(cmd *cobra.Command, args []string) error {
+			// Init logger
+			if quietFlag {
+				logLevel = logLevelQuiet
+			}
+
+			InitLogger(logLevel)
```
caarlos0, line 55: here we should also probably set the output to `cmd.ErrOrStderr`

logger.go
```diff
@@ -0,0 +1,50 @@
...
+	logLevelQuiet LogLevel = 2
+)
+
+var (
+	logger    *log.Logger
+	loggerOut io.Writer
+)
+
+// InitLogger configures logger level
+func InitLogger(level LogLevel) {
```
caarlos0, line 25: this dont need to be exported

logger.go
```diff
@@ -0,0 +1,50 @@
...
+	logger = log.New(os.Stderr, "", 0)
+
+	SetLogLevel(level)
+}
+
+// SetLogLevel modify log level
+func SetLogLevel(level LogLevel) {
```
caarlos0, line 32: this dont need to be exported

logger.go
caarlos0: also:
- this dont need to be exported
- could probably use `iota`

elmarsan: Yeah I think you're right.

One doubt about logger naming, is it something for all logger related? (file name, all the functions in logger.go)
Or are you referring to only LogLevel?

logger.go
elmarsan: agree

logger.go
elmarsan: agree

logger.go
caarlos0: everything in the logger.go file (including the filename itself)

elmarsan:
Hey @caarlos0 thanks for the feedback.
I'd like to be aligned with you before applying more changes. 
So after reading your comments I've got couple of doubts.

Things I've done so far (related to your feedback):
1. Use iota in `LogLevel` consts
2. Remove expored members of `logger.go` (`initLogger` and  `setLoggerLevel`)

I need more details about:
1. Logger naming. I believe logger is a proper name but maybe it's about taste.
2. Way we initialize logger. You mentioned to pass `cmd.ErrOrStderr` to `initLogger` func. It should not be a big deal but changes a little bit the way we're approaching the problem. 
3. --quiet flag, I think we can't set it as verbose as default because it's a boolean flag. I might be missing something about this..

elmarsan:
Hey @bashbunni @caarlos0 

It's being a while since the last comment. I'd like to going forward with this pull request.
Could we discuss the details in order to finish with this feature, please?

bashbunni:
> 2. Way we initialize output. You mentioned to pass cmd.ErrOrStderr to outputLogger func. It should not be a big deal but changes a little bit the way we're approaching the problem.

I believe this is mainly for consistency. We're using `cmd.InOrStdin()` in our `main.go` already to initialize our program input.
 
> 3. --quiet flag, I think we can't set it as verbose as default because it's a boolean flag. I might be missing something about this...

We could check if quiet is true, if so, don't print outputs from logger, else print. We should only need a single boolean flag as no flag will be verbose. Let me know if you'd like more details here

caarlos0:
> 3\. --quiet flag, I think we can't set it as verbose as default because it's a boolean flag. I might be missing something about this..

if it is false, hence not quiet, it is verbose... 

I'll push some changes and open a PR to your fork

elmarsan:
> [elmarsan#2](https://github.com/elmarsan/vhs/pull/2)

@caarlos0 thanks a lot, you've simplified it a lot, I believe right now it's more clear and easy to understand.

elmarsan:
@caarlos0 @bashbunni that's ready for the final review I think

record.go
```diff
@@ -85,8 +85,8 @@ func Record(cmd *cobra.Command, args []string) error {
 
 	// We'll need to display the stdin on the screen but we'll also need a copy to
 	// analyze later and create a tape file.
-	var tape = &bytes.Buffer{}
-	in := io.MultiWriter(tape, terminal)
+	tape := &bytes.Buffer{}
+	in := io.MultiWriter(tape, nil)
```
maaslalani, line 89:
```suggestion
	in := io.MultiWriter(tape, terminal)
```

caarlos0 (approved):

maaslalani:
Rebased and merged in #189

PR #241, opened 2023-02-26
refactor: unify publish instructions

elmarsan:
Hey I've done a little refactor in the way publish instructions are displayed.

While I was working on this #222 I noticed an error with `--quiet` flag. (See attached imaged)
The `publishShareInstructions` function was not printing last link, with this change we have a better encapsulation.

P.D: Right now this is working fine and is not a problem. However I could be a problem when we'll merge the pull request mentioned above.

quiet-flag-missing-link

### DIFF
--- a/main.go
+++ b/main.go
@@ -118,7 +118,6 @@ var (
 					publishShareInstructions(url)
 				}
 
-				fmt.Println(URLStyle.Render(url))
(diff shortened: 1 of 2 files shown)


elmarsan:
After discuss with @caarlos0 and @bashbunni we've decided publish instructions will not be printed with `--quiet` mode.

Therefore we no longer need this changes.

PR #244, opened 2023-03-02
fix: split start into new function

caarlos0:
- moved the startup code to a `start` function
- `launcher.Lookpath` is ok to return empty, rod downloads the browser if `bin` is empty
- `StartTTYD` did not start `ttyd`
- do not eat `ttyd` startup errors


maaslalani:
Thanks so much Carlos, works great!

PR #245, opened 2023-03-03
fix: ttyd --once

caarlos0:
From the help:

>     -o, --once              Accept only one client and exit on disconnection

This might be useful if, for example, VHS gets killed for reason and can't clean up after itself, this will guarantee that the ttyd session is also finished, so no one can connect to a dangling ttyd afterwards.


maaslalani (approved):

maaslalani:
Thank you! This looks reasonable as long as it still works 😄 I'll give it a double check

caarlos0:
I did some testing, and seems like it does, but it never hurts to test in more systems :D

PR #261, opened 2023-03-19
feat: add alt key as new command

tomgroenwoldt:
This PR adds the left `Alt` key as a new command. This command works exactly like the already implement `Ctrl` command. I avoided naming the command `AltLeft` or similar, because the right `Alt` is widely known as `AltGr` so I guessed there would be no naming conflicts/misunderstandings.

The purpose of adding this command is the following:
I write tapes for the editor [helix](https://helix-editor.com/) and some its shortcuts consist of the `Alt` key.


tomgroenwoldt:
Thank you for the feedback. Do you have any complaints about this PR? The only problem I see right now is that the `Alt` key can't be recorded right now, only passed in via a `.tape` file. I will take a look into that next, if you are fine with that.

maaslalani:
That's okay for now! It would be awesome if the <kbd>Alt</kbd> key could be recorded. This PR looks great, I'll do a bit more testing and merge soon.

maaslalani:
Thank you so much for the awesome work here @tomgroenwoldt. I will add the `Alt` key to `tree-sitter-vhs` as well.

PR #262, opened 2023-03-19
add `nushell` to the list of supported shells

amtoine:
hello guys :wave: :yum: 

i'm a [`nushell`](https://www.nushell.sh) user and i'm trying to record a session for the shell
```bash
'
Output out.gif

Set Shell nu
Set TypingSpeed 100ms

Sleep 500ms
Type "ls | where type == file and size > 10Kb | get name"
Sleep 500ms
Enter
Sleep 5s
' | vhs
```

> **Note**
> this is all written in `nushell` :+1: 

## problem
when i run this, i can see `bash` launching first, then run `nu` from `bash` and finally drop in `nushell`.

i had a look at `shell.go` and couldn't see `nushell` in the list of supported shells, so i thought i could add it to drop directly inside `nushell` :yum: 

 
- mirrors the other supported shells
- adds `nushell` which runs `nu --interactive`

## before landing
i'm not familiar with `go` at all, so if i'm missing something, if i should run some tests on my side, ..., please tell me :wink: 

### hope you'll like that :relieved: :tada:

### DIFF
--- a/shell.go
+++ b/shell.go
@@ -5,6 +5,7 @@ const (
 	bash       = "bash"
 	cmdexe     = "cmd"
 	fish       = "fish"
+	nushell    = "nushell"
 	powershell = "powershell"
 	pwsh       = "pwsh"
 	zsh        = "zsh"
@@ -60,5 +61,8 @@ var Shells = map[string]Shell{
 	},
 	cmdexe: {
 		Command: []string{"cmd.exe", "/k", "prompt=^> "},
+	},
+	nushell: {
+		Command: []string{"nu", "--interactive"},
 	},
 }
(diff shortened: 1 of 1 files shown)


amtoine:
thanks to @thomasgoulet, i've been able to test the feature...

and it works just as expected :tada: 

## details
i've run
```
go build
```
and then used the `./vhs` in the root of the project and the `bash` running `nu` before the actual script is gone :ok_hand: 

> **Note**
> we need to use `Set Shell nushell` and not `Set Shell nu` as i did above :+1:

amtoine:
i've changed the name of the shell to be used in `Set Shell` from `nushell` to the simpler `nu` in 4f67bfcd3162b693fa0678a5e1769a3c031708d7 :wink:

maaslalani (approved):
Thank you so much!

amtoine:
glad you like it :blush: :sparkles:

Issue #220, opened 2023-02-02
Feature Request: Source command


bakks:
Some food for thought: I'd like to have a bunch of `.tape` files for a project I'm working on which share some configuration commands, e.g. terminal size, font size, theme, etc. It's pretty common with scripting systems to be able to include other scripts of the same type within the current script, i.e. a command which executes another script at the current line. So my proposal is that VHS adds a command called `Source`, which takes an argument of another script file, executes the content of that file, then continues on.

Here's an example:
common.tape
```
Set FontSize 18
Set Framerate 30
Set Padding 10
```

demo.tape
```
Output demo.gif
Source ./common.tape
Type "echo 'Welcome to VHS!'" Sleep 500ms  Enter  
Sleep 5s
```

The content from `common.tape` would be included on line 2, no other impacts.

Thoughts on this? Feel free to close this out if a bad idea, missing the point, or already possible 😃 .

maaslalani:
Hey this is an awesome idea! This will definitely be implemented, I will close this since we are already tracking this in: #133.

bakks:
Ahh missed that other ticket but yes it's the same idea. Thanks for following up here!

Issue #133, opened 2022-11-02
Include other tapes


Airblader:
I would love to have some

```
Include foo.tape
```

command to modularize tapes a little bit. For example, one may require some setup (logging into an application), or want some "signoff phrase" at the end of a session, which are the same across different tapes.

maaslalani:
Hey this is a really good idea! I definitely want to have something like this.

Airblader:
> Fwiw we already have plans of adding Functions / Macros to the language!

Oh that is awesome. I didn't dare ask for that, but was about to.

> The Include makes things slightly more dependent on multiple files which is obviously less nice than having 1 file map to 1 output.

I feel the hesitation. In i3/i3 it took us years to be convinced to add an include directive, but it is such a quality of life change.

It really doesn't need to be fancy. Just a "replace this line with the contents of that file", basically.

maaslalani:
> It really doesn't need to be fancy. Just a "replace this line with the contents of that file", basically.

Totally, this basically gives you `Macro`s without needing to implement much new syntax since the whole file is essentially a Macro. I think I am sold 😄

melMass:
I came to open this exact issue, having a way to share a base setup when documenting many independent gifs would remove a lot of copy paste :)

bakks:
Plus one for this feature request!

elmarsan:
Hey @maaslalani I'd love to work on this one.

I've got a few ideas about how to approach it but I believe would be great to have some guidelines. 
How would you do it? For example a new command that executes another tape? Like @bakks proposed in #220.

maaslalani:
Hey @elmarsan, awesome that you're interested in this issue! My thinking is that you have a command (let's call it `Include` or `Source`) that you can use to reference a `.tape` file (from the relative path of the VHS invocation or tape file):

```
Include ./foo.tape
```

I think the only things I would expect to have are that it shows me that the command currently running is being sourced from a different file, and if there are any runtime or parsing errors it tells me that it is from that file (and the line number) which makes it a bit harder than simply copying and pasting the contents of the file into the master tape file which would skew all of the line numbers.

elmarsan:
> Hey @elmarsan, awesome that you're interested in this issue! My thinking is that you have a command (let's call it `Include` or `Source`) that you can use to reference a `.tape` file (from the relative path of the VHS invocation or tape file):
> 
> ```
> Include ./foo.tape
> ```
> 
[...]

PR #276, opened 2023-04-01
Feature/source command

elmarsan:
This pull request adds new `source` command for include tape commands into other tape.
It's based on #133 and #220 

### Example

demo.tape 
```
Output ./tapes/demo.gif

Set FontSize 32
Set Width 1200
Set Height 600
Set LoopOffset 50%

Sleep 1s

Source ./tapes/external.tape

Type "Thanks for the help external tape :)"
```

external.tape
```
Type "echo 'I'm an external tape'" Enter
Enter
Sleep 1s

Type "Goodbye friend"
Enter
Sleep 1s
```

Screenshot from 2023-04-01 16-56-47

### Wrong example

Screenshot from 2023-04-01 17-00-22

### DIFF
--- a/command.go
+++ b/command.go
@@ -3,12 +3,15 @@ package main
 import (
 	"encoding/json"
 	"fmt"
+	"io"
+	"os"
 	"os/exec"
 	"strconv"
 	"strings"
 	"time"
 
 	"github.com/go-rod/rod/lib/input"
+	"github.com/mattn/go-runewidth"
 )
 
 // CommandType is a type that represents a command.
@@ -37,6 +40,7 @@ var CommandTypes = []CommandType{ //nolint: deadcode
 	TAB,
 	TYPE,
 	UP,
+	SOURCE,
 }
 
 // String returns the string representation of the command.
@@ -95,7 +99,12 @@ func (c Command) String() string {
 
 // Execute executes a command on a running instance of vhs.
 func (c Command) Execute(v *VHS) {
-	CommandFuncs[c.Type](c, v)
+	if c.Type == SOURCE {
+		ExecuteSourceTape(c, v)
+	} else {
+		CommandFuncs[c.Type](c, v)
+	}
+
 	if v.recording && v.Options.Test.Output != "" {
 		v.SaveOutput()
 	}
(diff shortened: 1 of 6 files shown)


maaslalani:
Hey @elmarsan, this is really awesome work. I'll try and review / polish this up soon.

One thing to think about / test is if you can have a `Source` command within a `Source`d file. I think it would be fine to not allow this if it's tricky but we should provide proper error messages and such to the user.

Otherwise if we do plan on supporting nested sourcing it should likely indent one level deeper. Also, we should have some mechanism to ensure you can perform a cyclic source otherwise you would get infinite recursion.

elmarsan:
> Hey @elmarsan, this is really awesome work. I'll try and review / polish this up soon.
> 
> One thing to think about / test is if you can have a `Source` command within a `Source`d file. I think it would be fine to not allow this if it's tricky but we should provide proper error messages and such to the user.
> 
> Otherwise if we do plan on supporting nested sourcing it should likely indent one level deeper. Also, we should have some mechanism to ensure you can perform a cyclic source otherwise you would get infinite recursion.

Hey @maaslalani thanks!

With the current code you cannot use a `Source` command within a `Source` file. I decided to follow this approach for simplicity.
It might be useful allow nested `Source` thus we need to take a decision.
Let me know what you prefer and I can do it.

maaslalani:
I think for simplicity that's good that we don't allow double nesting of `Source` commands 👍

elmarsan:
> I think for simplicity that's good that we don't allow double nesting of `Source` commands +1

Cool in that case, Iĺl add more tests to make sure everything is correct.

spatecon (approved):

elmarsan:
Hey @maaslalani  sorry, it's been a while since we last spoke.

I've added a test to check nested `Source` commands and simplified the existing ones.
Let me know whether the tests are okay or we should add more. 
Thanks!

maaslalani:
Hey @elmarsan, thanks so much for this PR, again!

Very excited to get this merged, I will play around and do some more manual testing soon!

maaslalani:
Hey @elmarsan, mind giving me push access to your repository? I've made some changes and rebased the PR and it's ready to merge!

elmarsan:
> Hey @elmarsan, mind giving me (force) push access to your repository? I've made some changes and rebased the PR and it's ready to merge!

Sure, I've sent you and invitation. Once accepted I'll give you god mode right xD.

maaslalani:
Awesome! I've pushed the latest changes!

maaslalani:
Thanks so much @elmarsan for the awesome work! This is a huge upgrade to VHS ❤️

PR #277, opened 2023-04-04
docs: Added ColorfulRight and RingsRight

rm-dr:
Currently, the documentation only mentions two of the four possible `WindowBar` styles. This PR fixes that.

It is true that the `*Right` styles will be used less than the default styles, but we probably still want to mention them in the docs.


maaslalani:
Thanks so much!

PR #195, opened 2022-12-16
feat: env var to optionally set go-rod no-sandbox

brumhard:
This PR adds an environment variable to enable [go-rod's no-sandbox option](https://github.com/go-rod/rod/blob/611fbdd4e29fedc18478586057ea293847da317d/lib/launcher/launcher.go#L228).

This was suggested by the go-rod maintainer in the following issue: https://github.com/charmbracelet/vhs/issues/45#issuecomment-1296121300

I know the issue is closed but I couldn't find any other way to integrate vhs in a docker build process like for example with the following Dockerfile:
```Dockerfile
# syntax=docker/dockerfile:1.4

FROM ghcr.io/charmbracelet/vhs
COPY <<EOF hello.tape
    Output hello.gif
    Type "echo test"
    Sleep 500ms
    Enter
    Sleep 1s
EOF
RUN vhs hello.tape
```

When running with this patch it works when setting the newly added environment variable:
```Dockerfile
# syntax=docker/dockerfile:1.4

FROM ghcr.io/brumhard/vhs:0.0.4
ENV VHS_NO_SANDBOX="true"
COPY <<EOF hello.tape
    Output hello.gif
    Type "echo test"
    Sleep 500ms
    Enter
    Sleep 1s
EOF
RUN vhs hello.tape
```

PR #284, opened 2023-04-12
Set VHS_NO_SANDBOX by default

Xiol:
Chromium won't start in the existing Docker container because it's trying to start as root, and it requires the --no-sandbox flag setting.

```
could not launch browser: [launcher] Failed to get the debug url: find: '/root/.config/chromium/Crash Reports/pending/': No such file or directory
[0412/083701.089501:ERROR:zygote_host_impl_linux.cc(100)] Running as root without --no-sandbox is not supported. See https://crbug.com/638180. 
```

This sets the flag that was added in https://github.com/charmbracelet/vhs/pull/195 by default.

Alternatively this could be added to the Docker command listed in the README (`-e VHS_NO_SANDBOX="true"`) or a non-root user configured in the container, but this seems to be the simplest route.


maaslalani:
I think this makes sense to have default in the `Dockerfile`, thanks so much! Really appreciated ❤️

Issue #287, opened 2023-04-15
Sleep 1m10s Invalid command: m10s


thingsiplay:
```
$ vhs --version
vhs version v0.4.0 (5409c97)
```

I am new to this program and installed it through pacman in Archlinux based distro. I have recorded my session with

```
$ vhs record > demo.tape
```

and when I try to playback it with

```
$ vhs demo.tape
```

it refuses to play with error message:

```
 97 │ Sleep 1m10s
             ^^^^ Invalid command: m10s
```

When I edit the file and change it to `Sleep 1s`, then it plays and creates it back fine.

maaslalani:
Hey there! VHS doesn't support multiple units of time currently (but we probably should in the future) for now instead of `1m10s` please use  `70s`!

thingsiplay:
I understand and everything worked fine after the manual intervention. But vhs was the program that wrote the script and 1m10s.

tr3mor:
This is because recording just writes time.Duration but parser supports only "s" and "ms". The easiest fix is to keep it in sec when it goes above 60,  here is PR #300 .

maaslalani:
This is fixed by #300, thanks again @tr3mor for the fix and @thingsiplay for reporting this issue ❤️

PR #300, opened 2023-05-05
fix: record sleep only in supported time units

tr3mor:
Fix the issue described in #287 when record writes duration in unsupported units by keeping it in seconds if sleep was more than 1min

### DIFF
--- a/record.go
+++ b/record.go
@@ -158,7 +158,7 @@ func inputToTape(input string) string {
 		// characters.
 		if TokenType(lines[i]) == SLEEP {
 			sleep := sleepThreshold * time.Duration(repeat)
-			if sleep >= 60*time.Second {
+			if sleep >= time.Minute {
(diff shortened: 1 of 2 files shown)


record.go
```diff
@@ -158,7 +158,11 @@ func inputToTape(input string) string {
 		// characters.
 		if TokenType(lines[i]) == SLEEP {
 			sleep := sleepThreshold * time.Duration(repeat)
-			sanitized.WriteString(fmt.Sprintf("%s %s", TokenType(SLEEP), sleep))
+			if sleep >= 60*time.Second {
```
maaslalani, line 161:
```suggestion
			if sleep >= time.Minute {
```

tr3mor:
Really love the tools you are building, guys, so happy to contribute. I have committed your suggestion.

PR #304, opened 2023-05-15
refactor/Hide-Show

### LINKED ISSUE #130: `Hide` is not working
I just noticed that `Hide` is not working on my machine using docker. The README looks like it too:

image

elmarsan:
This pull request adds two new commands, `Pause` and `Resume`.

All commands between `Pause` and `Resume` will be executed but not include into the output gif.
In order to achieve this behavior I did the following:

- Create new terminal in 'hidden mode', every time `Pause` command is executed, all the commands will be processed into new instance of ttyd.
- Encapsulate Page, browser, textCanvas, cursorCanvas, tty and close `VHS` fields into new `Terminal` struct.
- `VHS` now has `currentTerm`, `mainTerm` and `hiddenTerm` which are pointers to Terminal instances.
  -  `mainTerm`: Styling is only applied to this instance.
  - `hiddenTerm`: Simple ttyd terminal with the default configuration.
  - `currentTerm`: It's a mirror of `mainTerm` when execution is not pause and mirror of `hiddenTerm` when execution is paused.

## Demo

```Elixir
Output pause-resume.gif

Pause

Type "mkdir folder"
Enter

Resume

Type "echo 'See secret folder'"
Sleep 1s
Enter

Type "ls | grep folder"
Sleep 1s
Enter

Sleep 2.5s
```

Output
<picture>
  Example of pause and resume
</picture>

Closes #130

### DIFF
--- a/command.go
+++ b/command.go
@@ -303,7 +303,7 @@ func ExecuteSetFontSize(c Command, v *VHS) {
 // ExecuteSetFontFamily applies the font family on the vhs mainTerm.
 func ExecuteSetFontFamily(c Command, v *VHS) {
 	v.Options.FontFamily = c.Args
-	_, _ = v.mainTerm.Page.Eval(fmt.Sprintf("() => term.options.fontFamily = '%s'", c.Args))
+	_, _ = v.mainTerm.Page.Eval(fmt.Sprintf("() => term.options.fontFamily = '%s'", withSymbolsFallback(c.Args)))
 }
 
 // ExecuteSetHeight applies the height on the vhs.
--- a/examples/glow/glow.tape
+++ b/examples/glow/glow.tape
@@ -1,23 +1,26 @@
-Output examples/glow/vhs-glow.gif
-Output examples/glow/glow.ascii
+Output vhs-glow.gif
+Output glow.ascii
 
+Set Framerate 40
 Set Width 1600
 Set Height 1040
 
 Sleep 1s
 
 Type "glow"
-
 Sleep 100ms
-
 Enter
 
 Sleep 1s
 
 Hide
-Tab
+Sleep 5s
+Show
+
+Hide
 Type "/artichoke"
 Enter
+Sleep 2s
 Down 2
 Show
 
@@ -37,6 +40,7 @@ Up@400ms 5
 Hide
 Type "/ulysses"
 Enter
+Sleep 2s
 Show
 
 Sleep 0.5s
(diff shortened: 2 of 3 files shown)


elmarsan:
Hey, I'm having issues with the workflow. The tests are running fine on my local machine, but I suspect there might be a problem related to ttyd.

After reviewing the code, I noticed that I'm the first one using the `New()` method in the tests. This could be the root of the issues because the CI might not be ready for that yet."

maaslalani:
What are your opinions on making this the behaviour of `Hide` / `Show` instead of introducing new keywords `Pause` / `Resume`?

elmarsan:
@maaslalani I'm complete agree with that!

If we add these new commands, users will probably get confused. And could be hard to explain the difference in the documentation between `Hide`/`Show` and `Pause`/`Resume`.
Moreover I'm not sure if users find useful the current behavior of `Hide` and `Show`. It's weird to see how frames stack up and pop out all at once in the GIF after `Show` command.

But yes, I would just move this logic to `Hide`/`Show`

maaslalani:
Thanks so much @elmarsan, I'm excited for this one.

elmarsan:
@maaslalani is everything ok?

helpermethod:
What has happened to this PR? It would make VHS infinitely more useful!

elmarsan:
@maaslalani let me know if you need any help.

patniemeyer:
> What are your opinions on making this the behaviour of `Hide` / `Show` instead of introducing new keywords `Pause` / `Resume`?
> 
> I think this type of behaviour fits quite well with `Hide` / `Show` IMO. Since you are essentially performing hidden commands (on a different terminal).

I agree that this should be the default behavior of Hide / Show.  As a new user I was very confused when "Hide" did not actually seem to hide anything.  Breaking changes are always hard, but it won't get any easier later :) 

Maybe rename the old functionality to "fast forward" "start/stop"? :)

osterman:
@maaslalani would love to see this feature! It's very hard to set up the demo environment before recording, since [any rc files are disabled](https://github.com/charmbracelet/vhs/blob/f63afe6fd7d74cb46845889803bfcec410245e1c/shell.go#L24) (otherwise this would be a great alternative). The current behavior, showing all the pre-set up commands, detracts from the demo itself (using hacks like `Ctrl+L` still cause a flicker). Also, when wanting to record demos of specifically one aspect of a tool, all the steps to get to that part are irrelevant. I've resorted to using tools like `ffmpeg` to cut off the first part of the recording, but it's not ideal.

PR #314, opened 2023-05-26
fix #313

### LINKED ISSUE #313: the sleep command is executed in wrong time
https://github.com/charmbracelet/vhs/blob/8ce786b09c711f6c8362d2091cd58c75190b6e79/command.go#LL199C2-L199C2

it is supposed to be `time.Sleep(dur * time.Second)`

for example, in the now version, `sleep 2` will sleep for 2**ns**

english bad, hope you can understand me

luohuayouyi12138:


maaslalani:
Hey @luohuayouyi12138, thanks so much for this fix, however I believe things currently work correctly as we append the seconds unit (if it doesn't exist) in another part of the application:

See comment: https://github.com/charmbracelet/vhs/issues/313#issuecomment-1591374059

PR #325, opened 2023-06-09
tr3mor/main

maaslalani:
- feat: setting to control cursor blinking
- feat: create bool type
- feat: parser updated
- readme wording
- Update README.md
- use boolean
- fix: check type


parser.go
```diff
@@ -269,6 +269,17 @@ func (p *Parser) parseSet() Command {
 				)
 			}
 		}
+	case CURSOR_BLINK:
+		cmd.Args = p.peek.Literal
+
+		p.nextToken()
+
+		if p.cur.Type != BOOLEAN {
```
maaslalani, line 277: @tr3mor Would you mind changing this line in your PR?

maaslalani: Specifically we need to check `Type` (not `Literal`)

token.go
```diff
@@ -51,6 +51,7 @@ const (
 	NUMBER  = "NUMBER"
 	STRING  = "STRING"
 	JSON    = "JSON"
+	BOOLEAN = "BOOL"
```
maaslalani, line 54:
```suggestion
	BOOLEAN = "BOOLEAN"
```

examples/settings/set-cursor-blink.tape
```diff
@@ -0,0 +1,10 @@
+Output examples/settings/set-cursor-blink.gif
```
maaslalani, line 1: @tr3mor Can you please add this example to your PR? Thanks!

PR #327, opened 2023-06-13
add note on TypingSpeed for new users in `demo.tape`

johnstonmatt:
Added note to the comment that is generated when a user calls `vhs new` to call out that `TypingSpeed` can be customized globally, and defaults to 50ms


maaslalani:
Thanks so much @johnstonmatt!

Issue #322, opened 2023-06-07
Rendering braille characters without the empty dots


joshka:
**Describe the bug**
My terminal (iterm2) renders braille characters without the empty dots in the various fonts I have installed, but try as I might I can't seem to get vhs to render them the same:

![Braille.gif](https://vhs.charm.sh/vhs-5kahK780GMJXHWJ1IGiLPB.gif)

**Setup**
Please complete the following information along with version numbers, if applicable.
macOS / zsh / iterm

**To Reproduce**
```
Output braille.gif
Set Width 400
Set Height 300
Set Shell zsh
Type "echo ⢕"
Sleep 2s
Enter
Sleep 2s
```
Also attempted with `Set FontFamily FiraCode Nerd Font` and a bunch of other fonts including DejaVu

https://www.fileformat.info/info/unicode/char/2895/fontsupport.htm shows font support

**Expected behavior**

Empty dots are empty:

<img width="93" alt="image" src="https://github.com/charmbracelet/vhs/assets/381361/832b50c1-d945-4193-9d41-fb6f80c99093">

**Context**
We use braille characters to render dot images in TUIs (specifically [ratatui](https://github.com/tui-rs-revival/ratatui))

maaslalani:
Hey thanks for this issue! I have added a fallback for "Apple Symbols" which appears to me to fix this issue on macs. I will add the Linux fix shortly as well.

joshka:
Thanks! This should make rendering screencasts of [ratatui](https://github.com/tui-rs-revival/ratatui) just that little bit easier.

hasezoey:
at least for me on vhs 0.6.0 and Linux (Manjaro 23.0.0) and KDE Konsole, it renders braille correctly:

(using the vhs code from https://github.com/charmbracelet/vhs/issues/322#issue-1746812179)
![braille](https://github.com/charmbracelet/vhs/assets/10911626/347e40eb-f3a2-4787-b8dc-f2dac596cd27)

PR #329, opened 2023-06-15
feat: add Apple Symbols font fallback for braille characters

maaslalani:
This PR adds font fallback to Apple Symbols which allows braille characters to
be rendered without empty circles.


maaslalani:
Partially fixes #322

Issue #363, opened 2023-08-19
Flag for disabling rendering


cafkafk:
**Is your feature request related to a problem? Please describe.**
Over at https://github.com/eza-community/eza we would like to use vhs to do integration testing.

This is possible, but a problem we face is we're not aware of any way to disable generating gif's. We'd like to ideally have a flag that would disable gif generation entirely because:
- It slows down integration tests.
- It simply isn't necessary when we only use `Output <something>.txt`.

See https://github.com/eza-community/eza/pull/147 for details on usecase.

**Describe the solution you'd like**
Ideally, a flag for disabling generating and rendering a gif would be sufficient.

This has the advantage of being non-intrusive and breaking to most users, and allowing us to script an option into our build script to include that flag.

**Describe alternatives you've considered**
Another alternative would be to use a flag that takes a value, which would perhaps be a better solution for scripting like we are doing. But, as a user of this program, I'm not sure it's what I'd prefer UX wise, and I'd rather leave the extra effort to people writing scripts.

**Additional context**
If there is no one interested in making this, and a potential solution is just adding an option and a if statement somewhere, I could potentially create a PR for this. But I'm stretched a bit thin, and I've never really had a go at go, so I'd *prefer* not to.

I am aware of the github actions, but we'd really like to do it this way (while also using the github actions, probably).

Also, if there is another, easier and/or obvious way to do this, please let me know :p

drupol:
This would be also useful for #109

maaslalani:
Awesome! Thanks for considering using VHS for integration testing. Let me see if I can quickly disable GIF generation for situations when an output is set to text.

cafkafk:
> Awesome! Thanks for considering using VHS for integration testing. Let me see if I can quickly disable GIF generation for situations when an output is set to text.

for eza's use case at least, we've moved pretty far past this solution at this point

drupol:
I would definitely use that new feature.

PR #364, opened 2023-08-20
feat: add Set DisableRender option for testing.

PThorpe92:
Flag for disabling rendering #363 

This PR adds a "SET" option 

    set DisableRender <bool>

That does the obvious :smile: and keeps it from executing the ffmpeg command so just .txt output can be dealt more easily for our CI/CD testing plans over at [EZA](https://github.com/eza-community/eza) as can be seen at [#147](https://github.com/eza-community/eza/pull/147)

This was my first time looking at the code base, so please let me know if I missed or overlooked something. Also please let me know if/where you would like tests added for this behavior. All I added was to the default token/parser tests, and I couldn't figure out exactly where a general test for this would go, so please do let me know and I'll add it.
Everything I have tried with `Set DisableRender true` and with output being .txt has worked perfectly. 

Just want to say thanks for a great program, and everything you guys do is awesome :rocket: :rocket:

EDIT: Just ran into this again and figured I'd fix the conflicts.


mbechto:
I have a similar use case. Can we have this please @maaslalani? 😃

PThorpe92:
> I have a similar use case. Can we have this please ? 😃

Assuming you have the same use case (testing the output of CLI program), You might like to know that we found diffing the txt output to be less than deterministic unfortunately :unamused: 

We ended up going with [trycmd](https://github.com/assert-rs/trycmd) and are working on a program to generate the test cases.

What kind of output are you looking to test? It would prob work for something fairly simple but [eza](https://github.com/eza-community/eza) output got complex to test this way pretty quick

maaslalani:
Hey @PThorpe92, thanks for this PR, I think the concept makes sense to have but I would like to avoid increasing the surface area of the language by adding a new keyword for this feature. What do you think of simply modifying the `Output` to recognize an argument such as `"none"` or `"/dev/null"` or something along these lines:

```elixir
Output none
```

PThorpe92:
Yeah that makes sense to me, I'll make the adjustments this weekend :+1:

maaslalani:
Hey @PThorpe92, thanks for the awesome PR, we really appreciate it.

I think disabling renders makes sense to me. However, the approach I'd like to take is that if the `Output ascii.txt` is specified then we don't render a GIF. That way the user would then have to also specify another `Output out.gif` to actually render the GIF.

I will close this PR for now (just doing some spring cleaning 😄) but feel free to open a new PR if this feature is still needed with the mutually exclusive `Output` version.

PR #373, opened 2023-09-12
Fix VHS for Nushell version ^0.83

ClipplerBlood:
## Issue
Since Nushell version 0.83, the `let-env` command has been [removed](https://www.nushell.sh/blog/2023-07-25-nushell_0_83.html#breaking-changes).

This breaks VHS when recording `nu` tapes.

## Fix
Replace `let-env VAR` with `$env.VAR`

Also the --interactive command is redundant, since the --execute flag also starts the interactive mode


maaslalani:
Thanks so much @ClipplerBlood!

PR #377, opened 2023-10-03
Update themes

joshka:
This is 3 commits intentionally to make this PR easier to review:
- the first pretty prints themes.json to make the second diff easier to read
- the second updates the themes from the actual upstream location, not the backup file
- the third adds back in some fields that were removed from the second to make it easy to review the commit.

commit 28e710f9b8597b7f6536378a809182ac1bdd99af
Author: Josh McKinney <joshka@users.noreply.github.com>
Date:   Tue Oct 3 02:18:21 2023 -0700

    Add back previously removed theme fields
    
    In the previous commit, the cursor and selection colors were removed, as
    well as the credits field to make the commit easier to view. This commit
    adds them back.

commit 6059dfb56bba199283f1c16c69048c5a99d54661
Author: Josh McKinney <joshka@users.noreply.github.com>
Date:   Tue Oct 3 02:10:52 2023 -0700

    Update themes from upstream
    
    Change the update source to the source to the same one that
    windowsterminalthemes.dev uses rather than the backup theme file
    which is only used if that file cannot be loaded.
    
    https://github.com/atomcorp/themes/blob/master/app/src/App.tsx#L18
    
    Filter out the cursor, selection and credits fields from the source file
    to make the diff easier to grok (added back in the following commit).
    
    BREAKING CHANGES:
      - Gruvbox Dark renamed to GruvboxDark.
      - Monokai Octagon renamed to Monokai Pro (Filter Octagon)

commit f08737109fe71492a67d5cab3d8df910f600a4d3
Author: Josh McKinney <joshka@users.noreply.github.com>
Date:   Tue Oct 3 01:35:01 2023 -0700

    Make themes.json diffs easier to read
    
    This pretty prints the theme file rather than compressing it on to a
    single line. This makes it easier to see what has changed between
    versions.


maaslalani:
@caarlos0 I'm okay with merging this but my gut tells me we probably want to keep the JSON file minified for binary size purposes.

joshka:
> @caarlos0 I'm okay with merging this but my gut tells me we probably want to keep the JSON file minified for binary size purposes.

A quick check suggests that the file size isn't at all dominated by the whitespace changes in either the json or the built executable (on a Mac at least) and in fact doesn't even show up in du. I'm not super familiar with go executable slimming, but a few KB seems reasonable to me regardless of size constraints.

```
❯ git sw main
Switched to branch 'main'
Your branch is ahead of 'upstream/main' by 3 commits.
  (use "git push" to publish your local commits)

❯ go build

❯ du -h vhs *.json
 31M	vhs
248K	themes.json
 16K	themes_custom.json

❯ git sw update-themes
Switched to branch 'update-themes'
Your branch is up to date with 'origin/update-themes'.

❯ go build

❯ du -h vhs *.json
 31M	vhs
248K	themes.json
 16K	themes_custom.json
```

joshka:
In the latest push I fixed up the tests by removing the themes_custom.json which was full of duplicate themes from themes.json. I manually checked all these - all were exact copies (except sublette, which had short hex colors like `#ef7` in custom which were equivalent to long hex colors like `#eeff77` in themes.json.

Removed the code that loaded both theme files and fixed the theme count test.

caarlos0:
I think it wouldn't hurt to add `jq -r` to the makefile to minify the themes.

other than that, lgtm

joshka:
The pros to minifying don't outweigh the cons here. Storing the file unformatted makes it difficult to see the effect of any future PR and makes it difficult for developers to read and work with the json.

If we want to minify it, that would be best to do as part of the build, but even then I'd suggest only doing this if there's some measurable benefit in speed or size. Here the size difference is < 1%, and although I haven't checked the perf difference, it's likely similarly negligible (loading 100kb to RAM from a local file is pretty fast).

maaslalani:
Thanks @joshka, yeah I agree it's worth having the mini-fication on build.

We can likely achieve this through `//go:generate`. I think for now we can merge this and improve the file size afterward.

PR #386, opened 2023-10-08
chore: Fix lint on main

abhinav:
golangci-lint currently fails on main with:

```
command.go:6: File is not `goimports`-ed (goimports)
        "github.com/atotto/clipboard"
command.go:267:1: exported: exported function ExecuteCopy should have comment or be unexported (revive)
func ExecuteCopy(c Command, _ *VHS) {
^
command.go:271:1: exported: exported function ExecutePaste should have comment or be unexported (revive)
func ExecutePaste(_ Command, v *VHS) {
^
```

This fixes all three issues.

### DIFF
--- a/command.go
+++ b/command.go
@@ -3,7 +3,6 @@ package main
 import (
 	"encoding/json"
 	"fmt"
-	"github.com/atotto/clipboard"
 	"io"
 	"os"
 	"os/exec"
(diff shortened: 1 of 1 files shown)


maaslalani:
Thank you so much! We really appreciate the PR

PR #422, opened 2023-12-11
Allow Ctrl+Space

alexwforsythe:
I'd like to record tapes that can type Ctrl+Space to trigger a keybinding I have in tmux. I could achieve this by remapping the keybinding temporarily for VHS, but it was nearly as easy to just make this PR instead.

A similar change was made recently in https://github.com/charmbracelet/vhs/commit/b5f8be82aec8d538bd00149985fdb9dfa50e5d4f to allow Ctrl+Shift+Alt+Enter, so I figured this might also be acceptable.

### DIFF
--- a/parser_test.go
+++ b/parser_test.go
@@ -207,48 +207,63 @@ func TestParseTapeFile(t *testing.T) {
 }
 
 func TestParseCtrl(t *testing.T) {
-	t.Run("should parse with multiple modifiers", func(t *testing.T) {
-		tape := "Ctrl+Shift+Alt+C"
-		l := NewLexer(tape)
-		p := NewParser(l)
+	tests := []struct {
+		name     string
+		tape     string
+		wantArgs []string
+		wantErr  bool
+	}{
+		{
+			name:     "should parse with multiple modifiers",
+			tape:     "Ctrl+Shift+Alt+C",
+			wantArgs: []string{"Shift", "Alt", "C"},
+			wantErr:  false,
+		},
+		{
+			name:     "should parse with out of order modifiers",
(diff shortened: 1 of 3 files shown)


parser_test.go
```diff
@@ -207,48 +207,73 @@ func TestParseTapeFile(t *testing.T) {
...
+		},
+		{
+			name:     "should parse with multiple modifiers",
+			tape:     "Ctrl+Shift+Space",
+			wantArgs: []string{"Shift", "Space"},
+			wantErr:  false,
+		},
+		{
+			name:     "should parse with multiple modifiers",
+			tape:     "Ctrl+Space+Shift",
```
maaslalani, line 236: I think this should be an error as Space is a key and should not come before a modifier.

parser_test.go
```diff
@@ -207,48 +207,73 @@ func TestParseTapeFile(t *testing.T) {
...
+			wantArgs: []string{"Space", "Shift"},
+			wantErr:  false,
+		},
+		{
+			name:    "Ctrl+AltRight",
+			wantErr: true,
+		},
+		{
+			name:    "Ctrl+Backspace",
```
maaslalani, line 245: We should probably add Ctrl+Backspace as valid, but this makes sense for now!

maaslalani (approved):

parser_test.go
alexwforsythe: Done! Added in a new commit

parser_test.go
alexwforsythe: Done! Added in a new commit

maaslalani:
Thank so much for the contribution @alexwforsythe!

Issue #356, opened 2023-08-12
[Docs] mention how to escape characters in `Type`


hasezoey:
Currently i cannot find any documentation on how characters in `Type` could be escaped, for example escaping `"`, where i already tried `\"` but it says invalid character

if it should be added, it should maybe be added to [this section](https://github.com/charmbracelet/vhs#type)

example:
```vhs
# unescaped, error
Type "echo "Test""

# trying to escape, also error
Type "echo \"Test\""

# changing to single quotes works
Type "echo 'Test'"
```

jskulski:
I am trying to use vhs with an environment variable and not finding it straightforward. So agree that adding something to the docs about escaping would be very helpful. I'll post here if I figure it out.

alexwforsythe:
FYI support for backticks was added in #157, so you don't need to escape quotes if you use them:

```sh
Type `echo "$MY_ENV_VAR"`
Type `echo "I can even use 'mixed quotes'"`
```

maaslalani:
Thanks so much for the issue @hasezoey

Fixed by https://github.com/charmbracelet/vhs/pull/429#

hasezoey:
just to confirm, this means the still is not "proper" escaping support? (ie using the character that starts the string within the string)

PR #429, opened 2024-01-07
docs: Update README.md to include backticks and single quotes

donovanhubbard:
When using the `Type` command you can enclose the text with either a single quote, double quote, or backtick character. However, the readme only provides examples of the double quote character which leads to issues like https://github.com/charmbracelet/vhs/issues/356.

This fixes https://github.com/charmbracelet/vhs/issues/356 by adding an examples with backticks and single quotes.


maaslalani:
Thank you so much for the contribution @donovanhubbard, I will merge this PR but also make some edits to combine this documentation with the code block where `Type` occurs.

Issue #320, opened 2023-06-03
Set the CWD of the command


blacktop:
I suppose I could use the technique of running `cd DIR` and hiding that cmd from the output, but thought it's be nice to have a higher order `Set` for the CWD as well?

osterman:
Is this the same as #19? (which was unfortunately rejected)

osterman:
Also, #304 could solve this pain point by making it easier to run a series of set up commands that do not get recorded.

PR #432, opened 2024-01-13
feat: Set the shells starting directory

donovanhubbard:
This adds a feature that I've wanted and has been requested by others (#320) is to set the current working directory of the terminal. This is done by using the `Set CWD` options such as

```
Set CWD "/home/user"
```

Because of the way the parser interprets the `/` character the option must be quoted. Includes a check to make sure the directory is valid.

### DIFF
--- a/README.md
+++ b/README.md
@@ -262,6 +262,14 @@ Set the shell with the `Set Shell <shell>` command
 Set Shell fish
 ```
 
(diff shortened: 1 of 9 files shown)


osterman:
This is a great addition. Although if #304 were merged, it could also solve this pain point.

PR #454, opened 2024-03-08
Update 'README.md' File to Include Winget method of installing vhs

og-mrk:
### About the Install Command
The `--exact` & `--id` options that's used with `winget install` command to ensure that the Exact Package is to be installed, which's `charmbracelet.vhs` package

### References for further reading
1. **VHS's PR**: https://github.com/microsoft/winget-pkgs/pull/142995
2. **ttyd's PR**: https://github.com/microsoft/winget-pkgs/pull/142883

**Note**: The `vhs` package requires `FFmpeg` & `ttyd` packages to be installed as well, you can read the Installer's Manifest in the VHS's PR linked above.

### Final Notes
Any feedback well be appreciated 😄


bashbunni:
Hey @og-mrk thank you so much for this contribution! What happens exactly if we don't use the `--exact` and `--id` flags with winget?

We previously merged a similar change to the docs in [Gum](https://github.com/charmbracelet/gum/pull/401/files), but in that case those flags weren't used. Could you kindly advise? I've never used Winget before, so that would be helpful to know. It might also mean that we need to update these changes that got merged in other repos :sweat_smile:

We would also like to keep the formatting consistent across projects, soooo that's probably something that would make sense to address while we're at it :smile: 

Other Winget PRs:
https://github.com/charmbracelet/glow/pull/500/files
https://github.com/charmbracelet/gum/pull/401/files
https://github.com/charmbracelet/soft-serve/pull/304

Some of these have been merged already, but if we're changing formatting to make 'em all consistent, it would be great to see a new PR with the consistent formatting

og-mrk:
Hello @bashbunni, thanks for taking your time on reviewing this PR.
Now about installing Packages using winget... now adays you can do it in different ways, from what I know, here's the following ways of installing vhs package (the concepts are the same with other packages):

> [!NOTE]
> From what I can see.. the `install` & `search` commands (and I believe other commands as well) are case-insensitive by default, so you can do `winget install Git.Git` or `winget install git.git`, both of which will install the same package.

### Using the Moniker of Package
```
winget install vhs
```

> Note: other examples, such as `Git.Git` Package, has a Moniker of `git`, so you can do `winget install git` and it'll work
> this's noted in one of the PRs you've highlighted: https://github.com/charmbracelet/soft-serve/pull/304

### Using Package Name
```
winget install vhs
```
### Using Package ID (It's generally made up of `PackagePublisher.PackageName`)
``` powershell
# The following will do the same action, which's installing vhs
# winget will treat them always as lower-case versions
# in other words, Upper Case letters will be seen as Lower Case version, like `myString.ToLower()` in GO For example
winget install charmbracelet.vhs
winget install Charmbracelet.VHS
winget install CharmBracelet.VHS
```
### Using the `--exact` option (`-e` is a shorter version) with any of the above (Will Become Case-Sensitive)
```
# Will Work (because there's a Package with this Exact Letter-Casing)
winget install -e vhs
winget install -e charmbracelet.vhs

# Will Not Work (No Packages with the following examples, at the time of writing of course)
winget install -e VHS
winget install -e charmbracelet.VHS
winget install -e Charmbracelet.vhs
winget install -e CharmBracelet.vhs
```

</br>

### Final Notes
IMHO don't recommend going with Moniker approach of installing packages, because multiple Packages can have the Exact Moniker, which will be an ambiguous request from winget's side, therefore the User will be prompted to choose one of the Packages. [[1]](https://github.com/microsoft/winget-cli/discussions/1672#discussioncomment-1665516) [[2]](https://github.com/microsoft/winget-cli/issues/1662#issuecomment-957707465)

Using the `PackagePublisher.PackageName` approach is the best course of action, for 99% of use cases, it'll be very clear who (`PublisherName` Field) made/released this Package (`PackageName` Field).

I'll remove the options `--exact --id` from this PR, because it won't really affect the final result, which's installing vhs.

caarlos0 (approved):

bashbunni:
awesome, @og-mrk thank you for explaining that to me as well :)

Issue #419, opened 2023-12-07
How can I change Shell Prompt


kevinah95:
**Is your feature request related to a problem? Please describe.**
I want to change the shell prompt.

**Describe the solution you'd like**
I'd like to have the chance to change the start symbol of the shell prompt, for example:

```bash
λ Commands with lambda at start
λ pwd
Σ Commands with sigma at start
Σ pwd
```

**Describe alternatives you've considered**
Add a config variable like:

Set Prompt λ
or
Set Shell Prompt λ

**Additional context**
Example:

![hacienda](https://github.com/charmbracelet/vhs/assets/7193395/9b1a3965-20d6-44f1-8c0d-53d315d8bf82)

edoardottt:
+1, interested on this feature

maaslalani:
Hey there! Thank you for the issue, I believe we should probably add some sort of setting for the `PROMPT` or allow setting environments (or both!):

Proposed solution:

```elixir
Set Prompt "> "
```

Or:

```elixir
Env PROMPT "> "
Env SECRET "$(pass SECRET)"
```

maaslalani:
Currently the prompt is hardcoded in: https://github.com/charmbracelet/vhs/blob/ed89b8b81081828548896e30dbe3770921f5e92c/shell.go#L23

and the default shell is `bash`.

What version of VHS and `ttyd` are you using and which operating system?

ocervell:
Weird, for me the default shell is `bash-5.0$` ...

Running latest version of VHS with Ubuntu 20.04.6 LTS running in a Vagrant VM:
```
vagrant@ubuntu2004:~/secator$ vhs -v
vhs version v0.7.1 (537d03a)

vagrant@ubuntu2004:~/secator$ lsb_release -a
No LSB modules are available.
Distributor ID: Ubuntu
Description:    Ubuntu 20.04.6 LTS
Release:        20.04
Codename:       focal
```

maaslalani:
I see, @ocervell do you have anything in your `~/.login` / `~/.bash_login` out of curiosity?

We use a `--login` shell and I'm wondering if perhaps the prompt is being overridden there. If that's the conflict then we can perhaps override with VHS by setting the prompt after that.

ocervell:
Nope, the files don't even exist on my machine.

We have this in the ~/.bashrc:
```sh
if [ "$ASCIINEMA_REC" == "1" ]; then
    PS1='\$ '
[...]

PR #469, opened 2024-04-18
cmd: add support for `ENV` command

Delta456:
Implements https://github.com/charmbracelet/vhs/issues/419

image

```haskell
Output examples/demo.gif
Env PROMPT "〉"

Require echo

Set Shell "bash"
Set FontSize 32
Set Width 1200
Set Height 600

Sleep 5s
```

and also allows adding environment variables.

image

```haskell
Env TESTING "this works"

Type "echo $TESTING"
Enter
Sleep 3s
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -744,9 +744,9 @@ Paste
 `Env` command sets the environment variable via key-value pair.
 
 ```elixir
-Env KEY "the command is working"
+Env HELLO WORLD
 
-Type "echo $KEY"
+Type "echo $HELLO"
 Enter
 Sleep 3s
 ```
--- a/README.md
+++ b/README.md
@@ -221,6 +221,7 @@ There are a few basic types of VHS commands:
 * [`Screenshot`](#screenshot): screenshot the current frame
 * [`Copy/Paste`](#copy--paste): copy and paste text from clipboard.
 * [`Source`](#source): source commands from another tape
+* [`Env <Key> Value`](#env): set environment variables
 
 ### Output
 
@@ -738,6 +739,35 @@ Sleep 500ms
 Paste
 ```
 
+### Env
+
+`Env` command sets the environment variable via key-value pair.
+
+```elixir
+Env KEY "the command is working"
+
+Type "echo $KEY"
+Enter
+Sleep 3s
+```
+
+`Env PROMPT value` changes the command prompt to the given value.
+
+```elixir
+Output examples/demo.gif
+Env PROMPT "Σ"
+
+Require echo
+
+Set Shell "bash"
+Set FontSize 32
+Set Width 1200
+Set Height 600
+
+Type "echo 'Custom prompt'" Sleep 500ms  Enter
+
+Sleep 5s
+```
 
 ### Source
 
--- a/README.md
+++ b/README.md
@@ -751,23 +751,6 @@ Enter
 Sleep 3s
 ```
 
-`Env PROMPT value` changes the command prompt to the given value.
-
-```elixir
-Output examples/demo.gif
-Env PROMPT "Σ"
-
-Require echo
-
-Set Shell "bash"
-Set FontSize 32
-Set Width 1200
-Set Height 600
-
-Type "echo 'Custom prompt'" Sleep 500ms  Enter
-
-Sleep 5s
-```
 
 ### Source
 
--- a/command.go
+++ b/command.go
@@ -253,14 +253,6 @@ func ExecuteCopy(c parser.Command, _ *VHS) {
 func ExecuteEnv(c parser.Command, v *VHS) {
 	_ = os.Setenv(c.Options, c.Args)
 
-	if c.Options == "PROMPT" {
-		if v.Options.Shell.Env != nil {
-			v.Options.Shell.Env[0] = strings.ReplaceAll(v.Options.Shell.Env[0], ">", c.Args)
-		} else if v.Options.Shell.Command != nil {
-			last := len(v.Options.Shell.Command) - 1
-			v.Options.Shell.Command[last] = strings.ReplaceAll(v.Options.Shell.Command[last], ">", c.Args)
-		}
-	}
 }
 
 // ExecutePaste pastes text from the clipboard.
--- a/command.go
+++ b/command.go
@@ -252,7 +252,6 @@ func ExecuteCopy(c parser.Command, _ *VHS) {
 // ExecuteEnv sets env with given key-value pair.
 func ExecuteEnv(c parser.Command, _ *VHS) {
 	_ = os.Setenv(c.Options, c.Args)
-
 }
 
 // ExecutePaste pastes text from the clipboard.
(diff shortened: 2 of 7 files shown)


command.go
```diff
@@ -248,6 +249,20 @@ func ExecuteCopy(c parser.Command, _ *VHS) {
...
+	_ = os.Setenv(c.Options, c.Args)
+
+	if c.Options == "PROMPT" {
+		if v.Options.Shell.Env != nil {
+			v.Options.Shell.Env[0] = strings.ReplaceAll(v.Options.Shell.Env[0], ">", c.Args)
+		} else if v.Options.Shell.Command != nil {
+			last := len(v.Options.Shell.Command) - 1
+			v.Options.Shell.Command[last] = strings.ReplaceAll(v.Options.Shell.Command[last], ">", c.Args)
+		}
+	}
```
Delta456, line 263: I couldn't find a better way to set up during the evaluation time. This also feels very hacky. I would love to know a better way to do this.

command.go
```diff
@@ -248,6 +249,20 @@ func ExecuteCopy(c parser.Command, _ *VHS) {
 	_ = clipboard.WriteAll(c.Args)
 }
 
+// ExecuteEnv sets env with given key-value pair.
+func ExecuteEnv(c parser.Command, v *VHS) {
+	_ = os.Setenv(c.Options, c.Args)
```
Delta456, line 254: Need to also un-set these environment variables after they are used(?)

maaslalani: Does this affect the actual users session? I imagine it doesn't and will disappear after the recording is over. If it does persist after the session is done then we could first save the user environment and then set the environment to the `oldEnvironment` afterwards.

command.go
Delta456:
```go
package main

import (
	"fmt"
	"os"
)

// Supported shells of VHS
const (
	bash       = "bash"
	cmdexe     = "cmd"
	fish       = "fish"
	nushell    = "nu"
	powershell = "powershell"
	pwsh       = "pwsh"
	zsh        = "zsh"
)

// Shell is a type that contains a prompt and the command to set up the shell.
type Shell struct {
	Command []string
	Env     []string
}

var prompt = func() string {
	prompt := os.Getenv("PROMPT")
	if prompt == "" {
		return "> "
	}
	return prompt
}()

// Shells contains a mapping from shell names to their Shell struct.
var Shells = map[string]Shell{
	bash: {
		Env:     []string{fmt.Sprintf("PS1=\\[\\e[38;2;90;86;224m\\]%s \\[\\e[0m\\]", prompt), "BASH_SILENCE_DEPRECATION_WARNING=1"},
		Command: []string{"bash", "--noprofile", "--norc", "--login", "+o", "history"},
	},
	zsh: {
		Env:     []string{fmt.Sprintf(`PROMPT=%%F{#5B56E0}%s %%F{reset_color}`, prompt)},
		Command: []string{"zsh", "--histnostore", "--no-rcs"},
	},
	fish: {
		Command: []string{
			"fish",
			"--login",
			"--no-config",
			"--private",
			"-C", "function fish_greeting; end",
			"-C", fmt.Sprintf(`function fish_prompt; set_color 5B56E0; echo -n "%s "; set_color normal; end`, prompt),
		},
	},
	powershell: {
		Command: []string{
			"powershell",
			"-NoLogo",
			"-NoExit",
			"-NoProfile",
			"-Command",
			fmt.Sprintf(`Set-PSReadLineOption -HistorySaveStyle SaveNothing; function prompt { Write-Host '%s' -NoNewLine -ForegroundColor Blue; return ' ' }`, prompt),
		},
	},
	pwsh: {
		Command: []string{
			"pwsh",
			"-Login",
			"-NoLogo",
			"-NoExit",
			"-NoProfile",
			"-Command",
			fmt.Sprintf(`Set-PSReadLineOption -HistorySaveStyle SaveNothing; Function prompt { Write-Host -ForegroundColor Blue -NoNewLine '%s'; return ' ' }`, prompt),
		},
	},
	cmdexe: {
		Command: []string{"cmd.exe", "/k", fmt.Sprintf("prompt=^%s ", prompt)},
	},
	nushell: {
		Command: []string{"nu", "--execute", "$env.PROMPT_COMMAND = {''}"},
	},
}
```

Was trying something like this but it didn't work.

command.go
Delta456: Ok it doesn't affect and disappears.

Delta456:
Ready for review

command.go
maaslalani, line 263: Let's remove this hack for now and keep thinking how to solve `PROMPT`s

maaslalani, line 263:
```suggestion
```

Delta456: Sure

README.md
```diff
@@ -738,6 +739,35 @@ Sleep 500ms
...
+
+Set Shell "bash"
+Set FontSize 32
+Set Width 1200
+Set Height 600
+
+Type "echo 'Custom prompt'" Sleep 500ms  Enter
+
+Sleep 5s
+```
```
maaslalani, line 770:
```suggestion
```

README.md
```diff
@@ -738,6 +739,18 @@ Sleep 500ms
 Paste
 ```
 
+### Env
+
+`Env` command sets the environment variable via key-value pair.
+
+```elixir
+Env KEY "the command is working"
```
maaslalani, line 747:
```suggestion
Env HELLO WORLD
```

README.md
```diff
@@ -738,6 +739,18 @@ Sleep 500ms
...
+
+Type "echo $KEY"
```
maaslalani, line 749:
```suggestion
Type "echo $HELLO"
```

maaslalani:
Awesome work @Delta456, thank you so much!

command.go
```diff
@@ -248,6 +249,12 @@ func ExecuteCopy(c parser.Command, _ *VHS) {
 	_ = clipboard.WriteAll(c.Args)
 }
 
+// ExecuteEnv sets env with given key-value pair.
+func ExecuteEnv(c parser.Command, v *VHS) {
```
maaslalani, line 253:
```suggestion
func ExecuteEnv(c parser.Command, _ *VHS) {
```

command.go
```diff
@@ -248,6 +249,12 @@ func ExecuteCopy(c parser.Command, _ *VHS) {
 	_ = clipboard.WriteAll(c.Args)
 }
 
+// ExecuteEnv sets env with given key-value pair.
+func ExecuteEnv(c parser.Command, _ *VHS) {
+	_ = os.Setenv(c.Options, c.Args)
+
```
maaslalani, line 255:
```suggestion
```

rahji:
Can we get a new Release so this feature is included when someone does `go install github.com/charmbracelet/vhs@latest`? PS the helix editor's syntax highlighting file will need updating.

SVendittelli:
I second the request for a new release. I install using `winget` and I'd like to be able to use the new `Env` functionality.

PR #475, opened 2024-04-29
Release ShiftLeft in Shift+Tab

akarpovskii:
This fixes a typo in Shift+Tab command. The command was pressing ShiftLeft and releasing AltLeft.


maaslalani:
Thanks so much for catching this!

PR #478, opened 2024-05-08
record: set VHS_RECORD=true environment variable in process

rockorager:
Set the environment variable VHS_RECORD=true in the spawned pty command
to allow processes to detect if it is running in a real TTY or in a VHS
recording. Certain TUI libraries query the terminal for feature
detection which can cause issues when running within a VHS pty. Setting this environment variable allows processes to detect if they are running in VHS and enable/disable features as needed.


maaslalani (approved):
This makes sense to me, thank you so much!

Issue #164, opened 2022-11-17
display key *names*


xkcd386at:
Awesome project thank you.

What I would like to see is the key *names*, at least for non-printing keys (like Enter, Escape, etc.), so when a user sees a demo he sees across the bottom what keys are being typed.

I've seen some tools do this.  Apologies if I am not explaining it well.

donovanhubbard:
One way to implement this might be to add the feature to ttyd which is what vhs uses to render the terminal.

Utagai:
This is something I've wanted pretty badly for another project of mine where the keypresses are important to show in the recording. I've made some headway over [here](https://github.com/utagai/steno).

Utagai:
You can follow the work on this at [my branch](https://github.com/Utagai/vhs/tree/keypress-overlay).

Utagai:
@luccahuguet If you look at the linked PR, I put up a patch to try to get this merged into `vhs` a long time ago. I tried pinging folks but it seems like no one wants to actually review my PR or merge it, which is perfectly fair if that's the maintainer's preference. No hard feelings from me, but it does mean that I think you won't get a satisfactory answer.

For your purposes, you could try using that branch. Alternatively, you can try my personal project `steno`: https://github.com/utagai/steno that tries to do this in a different way.

CC @maaslalani I would still be happy to work with you to merge my PR. No problem if you don't have the bandwidth, or don't want it to be merged. If it is the latter though, perhaps we should close the PR with a comment.

vic:
Hey @Utagai I'm also in need for this feature, will look at your `steno` project, and thank you for all you did on the `PR` hopefully it could be merged.

pkazmier:
I’ve implemented subtitles for keystroke captions in PR https://github.com/charmbracelet/vhs/pull/719. You’ll find lots of examples showing the many configuration options available.

PR #479, opened 2024-05-14
fix: correct typo in ffmepg.go

Utagai:
Just something I noticed while working on #164. Let me know if I've missed anything, since I wasn't sure if the typo was intentional... I think if it is though there should be a comment explaining why at the top.

This is my first PR against `vhs`, so I'm sorry if I'm messed anything up. I read [contributing guidelines](https://github.com/charmbracelet/.github/blob/5be06999b2aaac7640317437a373b60099f2b756/CONTRIBUTING.md), but I wasn't sure if there were any tests besides what is run via `go test ./...`.


maaslalani:
Oh wow! Thank you, not sure how I missed that!

PR #480, opened 2024-05-18
Command err handling

mastercactapus:
Updates command functions to return errors when unable to continue.

- Many silently ignored errors are now reported
- Fixed a race condition where some early SET commands would fail to access `term` if it hadn't started yet.

Dependency of #257 to address the concern of stopping/bubbling up if an error occurs without resorting to panic https://github.com/charmbracelet/vhs/pull/257#issuecomment-1878088189

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,3 +8,4 @@ completions
 dist
 vhs
 .idea/
+.vscode/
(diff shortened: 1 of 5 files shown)


maaslalani:
This is fantastic @mastercactapus, thank you for this awesome work!

mastercactapus:
@maaslalani pushed a fix for the lint issue that came up

maaslalani:
Tested this out and it worked great. This is a great improvement to VHS, thank you so much!

Issue #468, opened 2024-04-18
Show what is being entered


MHNightCat:
**Is your feature request related to a problem? Please describe.**
When I want to demonstrate some more complex operations, I would like to have a display that can show the currently pressed keys for auxiliary instructions.

**Describe the solution you'd like**
Displays which keys are currently being pressed at the bottom or in a customizable location.

**Additional context**
[Example](https://github.com/charmbracelet/vhs/assets/107802416/6c01fad3-f0aa-4ddf-866b-04fd79f3b729)

Utagai:
This is something I've wanted pretty badly for another project of mine where the keypresses are important to show in the recording. I've made some headway over [here](https://github.com/utagai/steno).

P.S. After talking about it with someone on the Charm Discord, I decided that it should be done in a third program, rather than inside `ttyd` or `vhs`.

Utagai:
Well, that was quick! Sorry for the extra ping to anyone subscribed to this issue. I got the green light for attempting to put this into `vhs`, so I'll be hacking on moving my linked repo into `vhs` for now.

@maaslalani I'm thinking we either close this issue or #164, since I'm pretty sure they are duplicates.

maaslalani:
Closing as duplicate of #164

PR #496, opened 2024-06-30
feat: Add keystroke overlay

Utagai:
Adds functionality for injecting a keystroke overlay into a `vhs` recording. This addresses #164 & #468. A large portion of this work is adapted from my work on [steno](https://github.com/utagai/steno).

The main overview of the approach here is that:
- We add support for `Set KeyStrokes [Hide|Show]` in tape files. This can be toggled on/off throughout the recording.
- We wrap various types from `rod`, including `rod.Page`, `rod.KeyActions` and `rod.Keyboard`, and make it so that any time `vhs` codebase wants to "type" something in some fashion, it goes through these decorated wrappers. The wrappers will record keystrokes.
- At the "end", when we go to call `ffmpeg` to generate output files, we use a series of `drawtext` filters based on the recorded keystroke events to render the keystroke overlay.

## Open Questions
- [ ] I currently use a series of symbols to print non-printable characters (e.g. space, control, etc). These can sometimes lead to issues if the system's font configuration uses a font that does not have the glyph in question and renders tofu. The user can fix this via `fontconfig`, which is what `ffmpeg` uses, but I'm open to changing the code to only use basic ASCII for example.
- [ ] Should I regenerate all the example GIFs? They shouldn't really appear changed to the human eye, but the timing fix (73cf9f0dabffc1ce1d7df917997d993d42da94dc) technically affects all recordings.
- [ ] Should we continue to hardcode the overlay font size as the `defaultFontSize`, or should we match whatever the specified font size is?


pkazmier:
I've added full support for ASS subtitles to VHS in PR https://github.com/charmbracelet/vhs/pull/719. Please check out the many demos I've included in the PR.

Issue #504, opened 2024-07-20
VHS_NO_SANDBOX ENV option needs documented


shaunco:
After banging my head against the wall on the on this error:
```
could not launch browser: [launcher] Failed to get the debug url: [0719/182133.729582:FATAL:zygote_host_impl_linux.cc(127)]
No usable sandbox! Update your kernel or see https://chromium.googlesource.com/chromium/src/+/main/docs/linux/suid_sandbox_development.md 
for more information on developing with the SUID sandbox. If you want to live dangerously and need an 
immediate workaround, you can try using --no-sandbox.
```

I came across https://github.com/charmbracelet/vhs/issues/45 which is referenced by https://github.com/charmbracelet/vhs/pull/195 which was merged over 2 years ago, and added `VHS_NO_SANDBOX` but doesn't seem to be documented anywhere outside of that PR.

For server mode, env options are documented in the readme ( https://github.com/charmbracelet/vhs/blob/main/README.md#the-vhs-server )...

PR #505, opened 2024-07-20
docs(README): document `VHS_NO_SANDBOX`

Delta456:
Fixes https://github.com/charmbracelet/vhs/issues/504

### DIFF
--- a/README.md
+++ b/README.md
@@ -205,6 +205,7 @@ vhs serve
 * `VHS_UID`: The User ID to run the server as (current user's UID)
 * `VHS_KEY_PATH`: The path to the SSH key to use (`.ssh/vhs_ed25519`)
 * `VHS_AUTHORIZED_KEYS_PATH`: The path to the authorized keys file (empty, publicly accessible)
+* `VHS_NO_SANDBOX`: Run with no-sandbox (`false`)
 
 </details>
 


shaunco:
`VHS_NO_SANDBOX` is not specific to server mode and should be in a different section. Would also be helpful to clarify why/when you'd want to use this, perhaps something like:

>When running vhs on Linux as root or with SUID (common for WSL-based Linux on Windows), Chromium may refuse to load due to weakened sandbox capabilities. This issue can be resolved by setting the environment variable VHS_NO_SANDBOX=true, which allows Chromium to run with a weakened sandbox. This enables vhs to function without requiring a non-root or non-SUID account.

Delta456:
> `VHS_NO_SANDBOX` is not specific to server mode and should be in a different section. Would also be helpful to clarify why/when you'd want to use this, perhaps something like:
> 
> > When running vhs on Linux as root or with SUID (common for WSL-based Linux on Windows), Chromium may refuse to load due to weakened sandbox capabilities. This issue can be resolved by setting the environment variable VHS_NO_SANDBOX=true, which allows Chromium to run with a weakened sandbox. This enables vhs to function without requiring a non-root or non-SUID account.

How about a separate heading for this?

caarlos0:
> How about a separate section for this? Like running on Linux as root or with SUID

yeah I think that would work better 🙏🏻

Issue #503, opened 2024-07-14
New ENV command breaks SET command in tape file


rahji:
**Describe the bug**

If you add a line like `Env PROMPT '>'` in the tape file, it has to be *after* any `Set` commands or it will cause them to be ignored when `tape` is run. (They appear in STDOUT but grayed out, and its clear from the resulting gif that they were ignored).

**Setup**
Please complete the following information along with version numbers, if applicable.
 - OS: Windows 11 WSL2 (Ubuntu 22.04.4 LTS)
 - Shell: default
 - Terminal Emulator: Windows Terminal
 - Terminal Multiplexer: tmux

**To Reproduce**
1. Create a tape file containing the following and run with `tape`:
```
Output foo.gif
Env PROMPT "-"
Set Height 100
Type "creates a gif, but ignores 'Set Height'"
```
2. Create a second tape file containing the following and run with `tape`:
```
Output bar.gif
Set Height 100
Env PROMPT "-"
Type "The tape command will fail with this file since 100 is not a valid 'Height'"
```

**Source Code**
See above

**Expected behavior**
Both tape files should work properly. Instead, the first one will create `foo.gif` but will ignore the `Set Height` command. The second one will fail since the `Set Height` line is not being ignored (and contains an invalid value)

**Screenshots**
n/a

**Additional context**
This only occurs on the current build. The ENV command is not available in the latest release. @Delta456's #469 was merged after that.

Delta456:
In my PR I removed the functionality of `ENV PROMPT prompt` and it only allows to define env variables.

rahji:
`ENV PROMPT prompt` works - but only if it comes after the `SET` lines

Delta456:
I will investigate this and attempt to fix this.

Delta456:
After hours of investigation, this happens for all commands before `SET`. Perhaps all commands need to be written after setting the configuration of a tape.

maaslalani:
> After hours of investigation, this happens for all commands before `SET`. Perhaps all commands need to be written after setting the configuration of a tape.

Hey @Delta456, this is a documented behaviour of settings:
https://github.com/charmbracelet/vhs#settings

> Setting must be administered at the top of the tape file. Any setting (except TypingSpeed) applied after a non-setting or non-output command will be ignored.

Delta456:
Ah, I missed that. This needs to be told when using the program IMO. Also, this issue can now be closed.

maaslalani:
[...]

PR #508, opened 2024-07-26
feat: warn when using `SET` commands not being the top decl

Delta456:
This PR adds a warning when using the `SET` commands after the top declaration.

> Agreed, if a setting is ignored we should display a warning that this setting should appear at the top of the file!
> _Originally posted by @maaslalani in https://github.com/charmbracelet/vhs/issues/503#issuecomment-2241697926_


caarlos0 (approved):

bashbunni:
Just tweaked the wording a bit. Thank you so much for the contribution :) 
image

PR #518, opened 2024-08-31
feat: don't create 'out.gif' by default

nolta:
If a tape file only asks for text output, e.g., `Output demo.txt`, `vhs` also creates an `out.gif` by default.

This PR removes that behaviour, and only outputs GIF files if `Output <name>.gif` appears in the tape file.


caarlos0:
thanks!

PR #551, opened 2024-11-24
fix: Inherit properties set via Source

abbra:
When tapes sourced into each other, properties set in included tapes aren't affecting the configuration in the tape that sources them. Fix this by allowing to parse them all within the current commands context.

### DIFF
--- a/parser/parser.go
+++ b/parser/parser.go
@@ -123,7 +123,7 @@ func (p *Parser) Parse() []Command {
 			p.nextToken()
 			continue
 		}
-		cmds = append(cmds, p.parseCommand())
+		cmds = append(cmds, p.parseCommand(cmds))
 		p.nextToken()
 	}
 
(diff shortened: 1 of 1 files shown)


caarlos0:
Hi!

this PR doesn't seem to fix the issue.

Going with #567 instead. 

Thanks for your work on this though!

abbra:
#566 is not the same as #551 -- you closed the latter claiming it does not do anything but you haven't tested it.

abbra:
I cannot test #567 today, sorry, too much work at $dayjob

caarlos0:
> #566 is not the same as #551 -- you closed the latter claiming it does not do anything but you haven't tested it.

sorry, in your description is not very clear what is actually not working, I thought it was the `Set` commands. 

Could you clarify what's the difference?

abbra:
If I set say, font and colors, in the included tape, they aren't reflected in anything that is done after the sourcing.

For example, if I had two tapes, `inc/base.tape` and `my.tape`, where the latter includes the former, then with the original version running the `my.tape` would not use the settings from the `inc/base.tape`. 
```
$ cat inc/base.tape
Set Theme GruvboxDark
Set Shell "bash"
Set FontSize 24
Set Width 1200
Set Height 800

$ cat my.tape
Output "my.webm"

Source "inc/base.tape"

Type "# Authenticate as an admin first"
Enter
...
```

PR #567, opened 2025-01-04
fix source

### LINKED ISSUE #566: Source: format settings have no effect
**Describe the bug**

Settings like `Width`/`Height` from a sourced file aren't adopted.

**Setup**
 - OS: manjaro/sway
 - Shell elvish/bash
 - Terminal Emulator: foot
 - Terminal Multiplexer: none

**To Reproduce**
1. Create a tape with width/height settings
2. Source it in another tape
3. Record gif
4. See resolution differ from settings.

**Source Code**

```vhs
Require example

Set Theme "Snazzy"
Set FontSize 30
Set Width 1400
Set Height 300
Set Padding 0

Set Shell "bash"

Hide
Type@0.01 "source <(example _carapace bash)" Enter
Type@0.01 "clear" Enter
Show
```

```vhs
Source ../bash.tape

Output out/word.bash.gif
Output out/word.bash.ascii

Type "example " Sleep 1 Tab@1 2
Type "q" # exit `more`
Type "ac" Sleep 1 Tab@1

Sleep 2
```

**Expected behavior**

Adopt settings from sourced file.

**Screenshots**

- current

word bash_bug

- expected
word bash_fixed

**Additional context**

rsteube:
Expands the parsed `Source` command instead of re-parsing it later.

fix #566


caarlos0:
does this do the same as https://github.com/charmbracelet/vhs/pull/551 ?

rsteube:
Oh, didn't see that one. I think so, but let me see.
Looks like the logged prefix might get lost there though.

This one is a bit tricky, so better have a closer look.

parser/parser.go
```diff
@@ -722,12 +723,23 @@ func (p *Parser) parseSource() Command {
 	if len(srcErrors) > 0 {
 		p.errors = append(p.errors, NewError(p.peek, fmt.Sprintf("%s has %d errors", srcPath, len(srcErrors))))
 		p.nextToken()
-		return cmd
+		return []Command{cmd}
 	}
 
 	cmd.Args = p.peek.Literal
+	filtered := make([]Command, 0, len(srcCmds))
```
caarlos0, line 730: we're not adding all the src into filtered, so I think probably can be 

```suggestion
	var filtered []Command
```

caarlos0 (approved):
just tested it, seems to work! thanks for the pr!

i have one small remark only :)

bashbunni (approved):
Looks great, thank you!

caarlos0:
see also https://github.com/charmbracelet/vhs/pull/604

Issue #574, opened 2025-01-17
Add a way to have variable typing speed


wizardlink:
**Is your feature request related to a problem? Please describe.**
I'd like a way of having variable typing speed, a range for the interval between keystrokes, to attempt to make it feel more human.

**Describe the solution you'd like**
I want to be able to set a variable, i.e. `TypingSpeedVaried [250ms, 350ms]` and have the `Type` actions pick a number in that range when waiting before the next input.

**Describe alternatives you've considered**
You can `Sleep` before every keystroke, but that is a very tedious process.

_Thank you for the amazing tool you've given us by the way!_

J-C-B:
Hi, if you scroll to the bottom of the readme.....

https://github.com/charmbracelet/vhs?tab=readme-ov-file#set-typing-speed

```
Type something really slowly!
`Type@500ms "Slow down there, partner."`
```

Also....

Set Typing Speed

```
Set TypingSpeed 500ms # 500ms
Set TypingSpeed 1s    # 1s
```

https://github.com/charmbracelet/vhs?tab=readme-ov-file#type

prithvijj:
Ehh I took a first stab at it https://github.com/charmbracelet/vhs/pull/580 ,
Feel free to check it out, if that helps with what you had in mind!

wizardlink:
Tested it, awesome work! Thanks for the contribution, can't wait for it to be merged.

meowgorithm:
Hey everyone! This is actually _great one, however I'm going to convert it into a discussion (for further discussion) until we've decided to integrate it. In general, we're aiming to be as thoughtful as possible about introducing VHS features right now given our internal capacity.

If we do decide to go ahead with this we'll convert it back to an issue.

PR #580, opened 2025-02-05
feat: Add `TypeVariable` Command for Variable Typing Speed

prithvijj:
## Changes

- Adding in the `TypeVariable` command which types out the given letters in the string, in a variable typing speed,
- The typing speed is determined by setting the `TypingSpeedVariable` option, and providing a min and max typing speed range
- After every keystroke, it would wait randomly within the min and max typing speed range, such that it feels more human,

## Testing Notes
Created `prits.tape`
```
Output prits.gif

Require echo

Set Shell "bash"
Set FontSize 32
Set Width 1200
Set Height 600
Set TypingSpeedVariable 50ms 555ms

TypeVariable "echo 'Welcome to VHS!'" Sleep 500ms Enter
```

Ran `go run . prits.tape` which produced `prits.gif`

image

prits.gif showing the second echo command, has multiple pauses after each keystroke, feeling less robotic

prits

## Other Notes

- Related to https://github.com/charmbracelet/vhs/issues/574

- I considered using another `TypeVariable` command instead of integrating it within `Type` itself. I guess it's easier to keep things separate, instead of multiple options affecting the same `Type` command,
- [ ] Not sure how the `TypeVariable@` would work, would be good to get some direction. Does `TypeVariable@0.1s 2s "blah blah"` be okay
- [ ] I think I'm doing something wrong around `parseTypeVariable`, in the sense of not parsing the `@1s` for indicating speed (i.e. overrides of the speed), Would be cool to get some help/direction on that
- [ ] Is there anything else I'm missing?

### DIFF
--- a/command.go
+++ b/command.go
@@ -4,6 +4,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"io"
+	"math/rand"
 	"os"
 	"os/exec"
 	"regexp"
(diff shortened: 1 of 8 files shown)


meowgorithm:
Hey @prithvijj! Just a note that per https://github.com/charmbracelet/vhs/discussions/583#discussioncomment-12086619 we're going to hold off on this until we've given the feature the green light. Given our internal bandwidth we're aiming to be thoughtful about introducing new features given our internal capacity.

Also, one question is the idea here that the two numbers in a directive like `Set TypingSpeedVariable 50ms 555ms` are the high and low values?

prithvijj:
> Hey @prithvijj! Just a note that per [#583 (comment)](https://github.com/charmbracelet/vhs/discussions/583#discussioncomment-12086619) we're going to hold off on this until we've given the feature the green light. Given our internal bandwidth we're aiming to be thoughtful about introducing new features given our internal capacity.

Yepp all good! 

> Also, one question is the idea here that the two numbers in a directive like `Set TypingSpeedVariable 50ms 555ms` are the high and low values?

Correct! just providing a range of min and max typing speed such that we wait after each keystroke for a given randomized number within the range, i.e. `type e, wait 100ms, type c, wait 554ms, type h, wait 55ms etc.`

PR #591, opened 2025-02-28
render: Make ffmpeg an optional dependency

3v1n0:
As per commit 92db33b ffmpeg is not required anymore by VHS unless some rendering is enabled.

For example we only use VHS as a testing tool and we rely on text output only, so having the whole ffmpeg installed isn't required.

So do this check at rendering time instead of as a pre-requisite


caarlos0 (commented):
thanks for the pr!

PR #598, opened 2025-03-12
docs: add "Enter" to cleanup example

mdjohns:
Hey all, 

Big fan of VHS (and Charm in general)!

Wanted to propose a small change to the "Cleanup" example.
I was scratching my head trying to figure out why my tape wasn't actually deleting the file, then realized it was missing the "Enter", so I thought it was worth clarifying in the readme.

### DIFF
--- a/README.md
+++ b/README.md
@@ -733,6 +733,7 @@ Enter
 # Cleanup
 Hide
 Type 'rm example'
+Enter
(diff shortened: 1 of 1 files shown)


raphamorim (approved):
Make sense to me, thanks for the PR!

PR #604, opened 2025-03-27
fix: source not respecting arguments of sourced commands

caarlos0:
refs #551
refs #567

it was setting the an empty string as argument to all sourced commands.


abbra:
LGTM! I tested it with my scenario and it works!

PR #605, opened 2025-03-30
fix(lexer): handle regex and escaped delimiters

### LINKED ISSUE #592: Impossible to escape / in Wait regex
**Describe the bug**

Using `Wait+Screen /foo\\/bar/` or something vhs fails with

```
22 │ Wait+Screen /foo\\\/bar/
        	            	                  ^^^^^^ Invalid regular expression 'foo\\\': error parsing regexp: trailing backslash at end of expression: ``
        	            	
        	            	 22 │ Wait+Screen /foo\\\/bar/
        	            	                          ^^^^ Invalid command: bar/
```

Or similar using a single-escape:

```
22 │ Wait+Screen /foo\/bar/
        	            	                  ^^^^ Invalid regular expression 'foo\': error parsing regexp: trailing backslash at end of expression: ``
        	            	
        	            	 22 │ Wait+Screen /foo\/bar/
        	            	                        ^^^^ Invalid command: bar/
```

The only way to check it is using `\x{2F}` instead of `/`

desertwitch:
fixes #592 

regex was parsed as string and not considering escaped delimiters
a regex sequence with an escaped delimiter would be cut short (see issue)

lexing logic for parsing regex was moved into a separate function
a regex sequence with an escaped delimiter is now properly handled
test cases were added to make this behaviour visible and testable for the future


desertwitch:
Missed a `.` at the end of a comment line, the linter didn't like that - but that's fixed and good to review now. 😉

lexer/lexer_test.go
```diff
@@ -95,6 +97,20 @@ Wait+Screen@1m /foobar/`
 		{token.NUMBER, "1"},
 		{token.MINUTES, "m"},
 		{token.REGEX, "foobar"},
+		{token.WAIT, "Wait"},
+		{token.PLUS, "+"},
+		{token.STRING, "Screen"},
+		{token.AT, "@"},
+		{token.NUMBER, "1"},
+		{token.MINUTES, "m"},
+		{token.REGEX, "foo\\/bar"},
```
raphamorim: This test expects "foo\\/bar" but the actual string stored should be "foo\/bar" after consuming the escape.

Issue #584, opened 2025-02-07
Get timestamps along with the video


imeckr:
**Is your feature request related to a problem? Please describe.**
I want to add subs to the video that is being generated by VSH, I would want to know when a particular command started and ended.

**Describe the solution you'd like**
Additional feature that returns timestamps for each command `vhs demp.tape --with-timestamps`

**Describe alternatives you've considered**
I tired approximating timestamps, but that approximation doesn't work for longer videos.

Soviut:
This would be really useful since I was hoping to apply sound effects to my videos and timecodes would make this much easier to insert.

Exar04:
Hey @Soviut @imeckr I made a pr based on your requested feature. Can you take a look at it and give a feedback on what could be improved in it.

PR #616, opened 2025-04-20
feat:get timestamp for each command with flat --with-timestamp or -t

Exar04:
Adds feature #584

Screenshot 2025-04-20 at 12 46 53 PM

### DIFF
--- a/main.go
+++ b/main.go
@@ -38,6 +38,8 @@ var (
 	publishFlag bool
 	outputs     *[]string
 
+	withTimestampFlag bool
+
 	quietFlag bool
 
 	//nolint:wrapcheck
@@ -255,6 +257,7 @@ func main() {
 
 func init() {
 	rootCmd.Flags().BoolVarP(&publishFlag, "publish", "p", false, "publish your GIF to vhs.charm.sh and get a shareable URL")
+	rootCmd.Flags().BoolVarP(&withTimestampFlag, "with-timestamp", "t", false, "Get timestamp for each command in the output")
 	rootCmd.PersistentFlags().BoolVarP(&quietFlag, "quiet", "q", false, "quiet do not log messages. If publish flag is provided, it will log shareable URL")
 
 	outputs = rootCmd.Flags().StringSliceP("output", "o", []string{}, "file name(s) of video output")
--- a/evaluator.go
+++ b/evaluator.go
@@ -6,6 +6,7 @@ import (
 	"io"
 	"log"
 	"os"
+	"time"
 
 	"github.com/charmbracelet/vhs/lexer"
 	"github.com/charmbracelet/vhs/parser"
@@ -137,6 +138,8 @@ func Evaluate(ctx context.Context, tape string, out io.Writer, opts ...Evaluator
 		}
 	}()
 
+	start := time.Now()
+
 	for _, cmd := range cmds[offset:] {
 		if ctx.Err() != nil {
 			teardown()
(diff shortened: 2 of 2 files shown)


main.go
```diff
@@ -255,6 +257,7 @@ func main() {
 
 func init() {
 	rootCmd.Flags().BoolVarP(&publishFlag, "publish", "p", false, "publish your GIF to vhs.charm.sh and get a shareable URL")
+	rootCmd.Flags().BoolVarP(&withTimestampFlag, "with-timestamp", "t", false, "Get timestamp for each command in the output")
```
raphamorim, line 260: I think is ok to leave as `timestamp` instead of `with-timestamp`

raphamorim:
Hey @Exar04 thank you for the PR, it's indeed a nice feature to add. Two things though, could you add tests about this functionality? and second would be nice `--timestamp` if it actually receive an unix POSIX for retrieve timestamp like `'%d/%m/%Y_%H:%M:%S:%N'` would return `20/04/2025_11:47:43:32`. It would allow people configure their timestamp in the way they would want to.

Soviut:
One issue with this is that there isn't a separate timecode for each typed letter. One use case for the timecode is being able to add audio and this wouldn't be enough to add audio to each keystroke.

Exar04:
I see @Soviut
seems hard to implement but I will give it a try.

@raphamorim yeah I will add tests at the end, after the complete feature is implemented.

raphamorim:
Perfect, please lemme know if need help with anything. Feel free to ping on my discord

Soviut:
> I see @Soviut seems hard to implement but I will give it a try.
> 
> @raphamorim yeah I will add tests at the end, after the complete feature is implemented.

You could infer the current typing speed which is how frequently the `Type` command types a character. Then multiply that by the number of characters.

https://github.com/charmbracelet/vhs?tab=readme-ov-file#set-typing-speed

The `TypingSpeed` changes the speed globally. While, you can also change the speed of a single `Type` command.

To handle timecodes with milliseconds it's usually `HH:MM:SS,mmm`. There's also `HH:MM:SS:FF` that will capture the current frame, assuming we know the framerate.

image

Exar04:
Screenshot 2025-04-22 at 1 04 03 PM
Does something like this works for you @Soviut ? 

Also what should be the names for the the tags @Soviut @raphamorim

Exar04:
> `20/04/2025_11:47:43:32`

I mean its just a stopwatch. should it have dates and stuff?

raphamorim:
Hey @Exar04 timestamp term isn't just about dates when comes to unix, but sure just allow formating time (hour, minute, second, millisecond) would be ok imho

Soviut:
> Does something like this works for you @Soviut ?

That's definitely more useful.

I think the ` : ` should be changed to something that does not appear in the timecode. Perhaps ` | ` or ` > `

> Also what should be the names for the the tags @Soviut @raphamorim

Do you mean the flags in the CLI?
I would go with:

```
--timing=timecode
--timing=timestamp
```

Where `timecode` produces `00:00:00.000` and timestamp produces milliseconds since the epoch.

Exar04:
Screenshot 2025-04-23 at 9 05 28 PM
Hey sorry, but I think I have not implemented the feature correctly. Theres a inconsistency in timecodes. I will have to take a deeper look into it later. Sry this might take longer for me to implement it

raphamorim:
No stress/rush @Exar04 please take your time

dzuelke:
Just a thought here: making the timecodes compatible with the FFmpeg metadata format (https://ffmpeg.org/ffmpeg-formats.html#Metadata-2) would allow for relatively easy tagging of videos with chapter markers (http://ikyle.me/blog/2020/add-mp4-chapters-ffmpeg or https://medium.com/@dathanbennett/adding-chapters-to-an-mp4-file-using-ffmpeg-5e43df269687 for details).

In fact, imagine `-t=chapters`, and a new `.tape` command like `Chapter`, that accepts a chapter name as the argument - the entire FFmpeg file could be auto-generated, and then the resulting `.mp4` has chapter markers!

PR #635, opened 2025-06-09
Add text screenshot capability to Screenshot command

wesen:
Extends the `Screenshot` command to support capturing terminal content as 
plain text in addition to the existing PNG image format.

## Usage

```
# Capture visual screenshot
Screenshot output.png

# Capture text content
Screenshot output.txt
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -756,12 +756,18 @@ Type "You will see this being typed."
 
 ### Screenshot
 
-The `Screenshot` command captures the current frame (png format).
+The `Screenshot` command captures the current frame in either PNG or text format.
 
 ```elixir
-# At any point...
+# Capture as PNG image
 Screenshot examples/screenshot.png
+
+# Capture as plain text
+Screenshot examples/screenshot.txt
 ```
+
+When using `.png` extension, VHS captures a visual screenshot of the terminal.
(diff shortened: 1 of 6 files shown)


joshka:
> Hey @wesen
> 
> can you elaborate on what's your use case for this? You got me curious :)

I'm not @wesen, but I can see significant value in this idea for the following 2 use cases:
- creating end user docs that are validated to be exactly the same as the output of some cli / tui app
- outputting values that can be used for snapshot testing for end-to-end testing situations

README.md
```diff
@@ -756,13 +756,19 @@ Type "You will see this being typed."
...
-# At any point...
+# Capture as PNG image
 Screenshot examples/screenshot.png
+
+# Capture as plain text
+Screenshot examples/screenshot.txt
 ```
 
+When using `.png` extension, VHS captures a visual screenshot of the terminal.
+When using `.txt` extension, VHS captures the terminal content as plain text without any styling.
```
joshka, line 770: Missing docs for .ansi

wesen: will add!

wesen:
@joshka @caarlos0 yes that's exactly my usecase. I use it very often for LLM agents to snapshot intermediate steps either for their own sake, or for me to verify they actually implemented what I wanted.

PR #672, opened 2025-10-17
fix: can't combine Shift and Down #641

### LINKED ISSUE #641: Can't combine Shift and Down
**Describe the bug**
I can't find a way to simulate the user holding shift and pressing down.

At fist I tried to implement this manually in the script, but nothing I tried worked. This is everything I tried and got compile errors:
```
Shift Down
Shift Down 1
Shift+Down
Key "shift" "down"
Key "shift+down"
```

I then tried to use `vhs record` to create what I needed. Performing a shift down in a recording session produces the following which does nothing when trying to record a demo using the script. 

```
Type "[1;2B"
```

**Setup**
 - OS: macOS
 - Shell: zsh
 - Terminal Emulator: iterm
 - Terminal Multiplexer: tmux

**To Reproduce**
Use the min example I put in the source code and run this vhs demo tape:

```
Output demo.gif

Set Framerate 10
Set Shell zsh
Set FontSize 24
Set Width 1200
Set Height 600

Type "go run main.go"
Sleep 1s
Enter
Sleep 2s

# Test shift down from vhs record:
Type "[1;2B"
Type "[1;2B"
Type "[1;2B"
Sleep 5s 
```

**Source Code**

```go
package main

import (
	"fmt"
	"log"
	"os"

	tea "github.com/charmbracelet/bubbletea"
)

type model struct {
	events []string
}

func (m model) Init() tea.Cmd {
	return nil
}

func (m model) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
	switch msg := msg.(type) {
	case tea.KeyMsg:
		switch msg.String() {
		case "q", "ctrl+c", "esc":
			return m, tea.Quit
		case "shift+down":
			m.events = append(m.events, "Got shift+down!")
		default:
			m.events = append(m.events, fmt.Sprintf("Got key: %q", msg.String()))
		}

		// Keep only last 10 events
		if len(m.events) > 10 {
			m.events = m.events[len(m.events)-10:]
		}
	}
	return m, nil
}

func (m model) View() string {
	s := "Shift+Down Key Event Test\n"
	s += "========================\n\n"
	s += "Press Shift+Down to test the key event\n"
	s += "Press 'q' or 'esc' to quit\n\n"
	s += "Recent events:\n"

	for _, event := range m.events {
		s += "  " + event + "\n"
	}

	return s
}

func main() {
	p := tea.NewProgram(model{events: []string{}})
	if _, err := p.Run(); err != nil {
		log.Fatal(err)
		os.Exit(1)
	}
}
```

**Expected behavior**
`Got shift+down!` should be printed three times, but instead every key in the Type statement is sent.

**Screenshots**
Add screenshots to help explain your problem.

**Additional context**

security-log:
Fixes #641 

Changes

Modified: command.go
Keep the Shift modifier held while typing navigation keys so `Shift+Arrow/Page/Home/End` fire the correct rod key codes.

Modified: parser/parser.go
Allow Shift commands to consume navigation tokens (Arrow, Page, Home, End) in addition to the existing literals.

Modified: parser/parser_test.go
Add coverage for parsing `Shift+Down` and `Shift+Tab`.

Modified: record.go
Map Shift+Arrow escape sequences, sort substitutions by length to avoid partial matches, and preserve Shift-prefixed commands during sanitization.

Modified: record_test.go
Verify that the recorder now turns Shift+Arrow escape sequences into `Shift+…` commands.

Why

Shifted navigation keys were emitted as plain text (`Down`, `Up`) or broken into `Escape` fragments, so we couldn’t replay selections made with Shift while recording or parsing tapes. These changes teach the recorder, parser, and executor to treat shifted navigation as first-class commands.

How to reproduce

Before the fix

```
go test ./record -run TestInputToTape/Shift+Arrow -count=1
--- FAIL: TestInputToTape (0.00s)
    record_test.go:...: want:
    Shift+Down
    ...
    got:
    Escape
    [1;2B
    ...
```

After the fix

```
go test ./record -run TestInputToTape/Shift+Arrow -count=1
ok   github.com/charmbracelet/vhs/record
```


davidpfarrell:
Hoping this will provide `SHIFT+ENTER` support, too? (looking at the code suggests yes)

GustavoCaso:
Any updates on getting this contribution review @caarlos0 I think this is a quite useful addition to VHS 😄

Issue #638, opened 2025-07-08
Document the defaults for each setting / command


joshka:
**Is your feature request related to a problem? Please describe.**
I always have to grep the source to work out what the defaults for each setting are (e.g. Font size, Family, Width, height)

**Describe the solution you'd like**
- In the README, make sure every setting indicates the defaults
- In the file created by `vhs new` document the defaults

hesreallyhim:
(I think) Duplicate: https://github.com/charmbracelet/vhs/issues/495

hesreallyhim:
@joshka prefer to close the earlier one i guess since I reference this in my PR

Issue #495, opened 2024-06-27
Show the defaults for each setting in the docs and the docs tape file


joshka:
**Is your feature request related to a problem? Please describe.**
Try finding the default font size from the docs... or any other random setting

**Describe the solution you'd like**
Add this to both the readme and the docs.tape generated from vhs

**Describe alternatives you've considered**
none

**Additional context**
none

PR #680, opened 2025-11-29
docs: add reference to default options list

hesreallyhim:
- [-] ~I have created a discussion that was approved by a maintainer (for new features).~ (docs-only)

I found myself looking through the code to find the default values for various Options settings, so I thought it would be helpful to provide the user with a list of these settings in the README. So as to make the smallest visible change to the existing docs, I added a link in the README to a new Markdown table I assembled (with LLM assistance) based off of the info in `vhs.go` and related files like `styles.go`.

If you want to "check my work" there is a file in https://github.com/charmbracelet/vhs/commit/037d43708101f0a424c0868955c39e1ce2559857 which contains a CSV file I built up to compile the information including line references for the different settings.

Thanks, this library is amazing! 

(Partially?) Fixes: #638 
(Partially?) Fixes: #495 (duplicate)


hesreallyhim:
> I think that ideally, the setting, description and default should be in the same table where possible.

I just did it this way for minimal disturbance to the main README.

joshka:
> I just did it this way for minimal disturbance to the main README.

The problem I had with this to start is that you have to look in two places for the information. This simplifies the two places, but it remains at two. The 'what is the default' question is now answered by:
1. look at readme
2. search 'default' in readme
3. click and read other doc, flipping back to readme for docs.

hesreallyhim:
@joshka maybe I could put the defaults table inside a disclosure element. I'm open to whatever, and totally respect your opinion but also want to hear from a codeowner 😉

PR #707, opened 2026-02-20
feat: add viewport scroll commands to tapes

### LINKED ISSUE #404: Support scrolling the terminal
**Is your feature request related to a problem? Please describe.**
Let's say we have a command that outputs a bunch of data to the terminal, some of which causes the terminal to scroll. I'd like some way to cause the terminal to scroll up to see the previous data in the gif.

My specific use case for this is in Ratatui (a rust TUI lib) testing that we can insert information before the current viewport (writable part of the screen for an app), and that this is in the correct position after the draw.

**Describe the solution you'd like**
Two commands ScrollUp N and ScrollDown N that scroll by a number of rows

**Describe alternatives you've considered**
I'm not sure if there is a generic way to do this with keyboard shortcuts (most apps will not pass PageUp/PageDown to the terminal in a way that's useful)

**Additional context**
Looking for less manual ways to test https://github.com/ratatui-org/ratatui/pull/596

joshka-oai:
## Summary
- add `ScrollUp` and `ScrollDown` tape commands for viewport row scrolling
- wire new commands through token lookup, parser command handling, and command execution via `term.scrollLines()`
- update README/manual/demo fixture text and expand lexer/parser/command-count coverage

## Rendered output

scroll smoke

Tape used:

```tape
Output "/tmp/scroll-smoke.gif"
Set Width 900
Set Height 500

Type "for i in $(seq 1 60); do echo line-$i; done"
Enter
Sleep 500ms

ScrollUp 12
Sleep 700ms
ScrollDown 6
Sleep 700ms
```

## Testing
- go test ./...

Fixes #404

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -6,3 +6,5 @@ on:
 jobs:
   lint:
     uses: charmbracelet/meta/.github/workflows/lint.yml@main
+    with:
+      golangci_path: .golangci.yml
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -30,9 +30,24 @@ linters:
       - text: '(slog|log)\.\w+'
         linters:
           - noctx
+      - text: "var-naming"
+        linters:
+          - revive
+      - text: "G705"
+        linters:
+          - gosec
+      - text: "G115"
+        linters:
+          - gosec
+      - text: "os/exec.Command must not be called"
+        linters:
+          - noctx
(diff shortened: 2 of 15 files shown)


joshka-oai:
ping

aymanbagabas:
Hi @joshka-oai, thank you for your patience and contribution. We're looking into why the checks are failing, and will merge your PR very soon.

Issue #691, opened 2025-12-24
bash PROMPT_COMMAND causes errors


felixge:
**Describe the bug**

The presence of a PROMPT_COMMAND environment variable is prone to causing errors when feeding a tape into VHS.

This is common when using tool like mise, zoxide and/or custom prompt commands. For example my PROMPT_COMMAND includes `_mise_hook;__zoxide_hook;generate_prompt;` - all of these commands are bash functions, so they are not accessible from a VHS session that didn't source my bashrc.

**Setup**

* GNU bash, version 5.3.3
* macOS 15.6

**To Reproduce**

On my machine:

```
go run . ./examples/commands/enter.tape 
```

![Image](https://github.com/user-attachments/assets/916b30f8-9eb9-4f29-b1e8-81392cbc8991)

This reproducer should work on any machine:

```
PROMPT_COMMAND=does-not-exist go run . ./examples/commands/enter.tape 
```

![Image](https://github.com/user-attachments/assets/760de726-f18b-46c3-ba3c-eedf6d0533fb)

**Expected behavior**

```
PROMPT_COMMAND= go run . ./examples/commands/enter.tape 
```

![Image](https://github.com/user-attachments/assets/9933552a-8b80-4e55-9b10-b83bbf68c5fb)

**Additional context**

Proposed fix: https://github.com/charmbracelet/vhs/pull/692

Issue #659, opened 2025-08-23
`Wait+Screen` regex fails to match text in the viewport after scrolling


runesoerensen:
### Describe the bug

When a `.tape` file generates enough output to cause the terminal to scroll, [the `Wait+Screen /regex/` command ](https://github.com/charmbracelet/vhs?tab=readme-ov-file#wait)fails to find text that is printed/currently visible on the screen.

It appears that instead of searching the visible lines on the screen, the command searches the lines at the top of the terminal's scrollback buffer. This makes it impossible to use `Wait+Screen` to match new command output once a terminal has scrolled.

This issue is related to https://github.com/charmbracelet/vhs/issues/657. and likely stems from the `VHS.Buffer` method, which appears to [not account for the scroll position](https://github.com/charmbracelet/vhs/blob/1b7db05a951df65e30eb4fdb9429711e89c7e511/testing.go#L70).

### Setup

 - OS: macOS (15.6.1)
 - Shell: `zsh`, `bash`, `nu`

### To Reproduce

1.  Create a file named `wait-screen-test.tape` with the following content.

    ```
    Set Height 300

    Type "seq 10"
    Enter
    Wait+Screen /9/

    Type "echo foo"
    Enter
    ```

2.  Run the tape file: `vhs wait-screen-test.tape`.

### Actual Behavior

The `vhs` command execution halts and exits with a timeout error because the regex only matches against output in the beginning of the terminal buffer.

```
% vhs wait-screen-test.tape     
File: wait-screen-test.tape
Set Height 300
Type seq 10
Enter 1
Wait Screen 9
failed to execute command: timeout waiting for "Screen 9" to match 9; last value was: > seq 10
1                                                                                             
2                                                                                             
3                                                                                             
4                                                                                             
5                                                                                             
recording failed
```

### Expected Behavior

The `Wait+Screen` command should match the regex by searching the terminal buffer visible on the screen. The tape execution should then continue.

```
% vhs wait-screen-test.tape
File: wait-screen-test.tape
Set Height 300
Type seq 10
Enter 1
Wait Screen 9
Type echo foo
Enter 1
```

Issue #611, opened 2025-04-17
Add a setting which creates an automatic sleep between commands


Soviut:
**Is your feature request related to a problem? Please describe.**

In my tape files, the majority of my code are `Sleep 1s` commands that I insert between all the other commands. This is to give the viewer a moment to see what's been typed or selected before hitting Enter, then another moment to see what happened after the command ran.

**Describe the solution you'd like**

It would be nice if there was a setting in the tape file that could add an automatic sleep/delay between commands (except actual Sleep commands).

I propose calling it `Gap` which would normally have a default of `0s`

```elixir
Gap 1s                 # set initial gap size

Type "hello world"     # delay 1s after typing
Enter                  # delay 1s after hitting Enter
Type "typed 1s later"  # delay 1s after typing
Enter                  # delay 1s after hitting Enter
Sleep 5s               # delay 5s (ignore gap delay here)

Gap 2s                 # change gap size

Type "gap changed"     # delay 2s after typing
Enter                  # delay 2s after hitting Enter
```

**Describe alternatives you've considered**

Currently, the only way seems to be adding lots of Sleep 1s commands.

**Additional context**

![Image](https://github.com/user-attachments/assets/7c688a9e-a75f-42ad-a982-b3ce5e872b62)

Issue #537, opened 2024-10-31
Instead of Sleep'ing, can the tape 'auto-advance' upon completion of the previous command?


paragchandra:
**Is your feature request related to a problem? Please describe.**
I've started using VHS to record demos of command-line utilities I've been developing, which has been working pretty well - so first off, thanks for this great utility. I guess the main 'complaint' I have is that I have to repeatedly try a command to see how long it might take in the worst case, so that I know how long to Sleep for, so that VHS captures all the output of the command.

**Describe the solution you'd like**
I think it would be more intuitive if VHS could just wait for a command to complete - and optionally exit on nonzero return code - before proceeding to the next command. Maybe that can't be directly supported by the existing 'Type' command, since that could be just to type text into a text editor or similar, so perhaps a different keyword would be needed.

**Describe alternatives you've considered**
I've looked at `asciinema`, which does operate in this sort of 'blocking' manner I've described, but I _really_ like the ability to essentially 'script' my demos ahead of time and then let VHS take care of the rest, rather than typing live while `asciinema` records.

**Additional context**
Hopefully the description above is clear enough, and if this is 'by design', that's fine too..

tobihans:
I agree with this too. If core team could guide me on what would need to be done, I'm more than happy to try a PR.

pmonks:
Pretty sure this is (yet another) duplicate of issue #70, which has a fix (PR #257) which hasn't been released yet (issue #550).

caarlos0:
Hey!

This should now have been fixed on latest VHS by using the `Wait` keyword: https://github.com/charmbracelet/vhs/releases/tag/v0.9.0

PR #708, opened 2026-02-20
feat: add AwaitPrompt command for shell prompt detection

MattieTK:
## Summary

Adds a new `AwaitPrompt` command that waits for the shell to render a new prompt before proceeding. This provides a reliable, timing-independent way to synchronise tape playback with command completion.

### Syntax

```
AwaitPrompt          # wait using default WaitTimeout (15s)
AwaitPrompt@30s      # wait with explicit timeout
AwaitPrompt@2m       # longer timeout for slow commands
```

## Relationship to `Wait`

Plain `Wait` (without `+Line` or `+Screen`) already does something similar – it defaults to `Wait+Line` with the pattern `/>$/`, which matches VHS's default `> ` prompt. For the common case of "run a command and wait for the shell to be ready", both commands achieve the same result.

The difference is in reliability. `Wait` matches visible text in the terminal buffer, which means:

- **False positives**: if a command outputs a line ending in `>` (compilation output, markdown, `git log`, etc.), `Wait` returns early before the command has actually finished
- **Custom prompts**: if the prompt doesn't end with `>`, plain `Wait` stops working unless the user also sets `WaitPattern`
- **Buffer issues**: `Wait+Screen` has a known bug with scrolled content (#659); `Wait+Line` is less affected but still reads from the buffer

`AwaitPrompt` avoids these problems by detecting an invisible marker in the raw terminal stream rather than matching visible text. It works regardless of what the prompt looks like or what the command outputs.

### Could `Wait` be reworked instead?

Rather than adding a separate command, the OSC 133 marker mechanism could potentially be folded into `Wait` itself – making plain `Wait` (with no regex) use prompt detection by default, while `Wait /pattern/`, `Wait+Line`, and `Wait+Screen` continue to use regex matching as they do today.

This would mean existing tapes that use bare `Wait` would silently become more reliable, with no syntax changes needed. We're happy to rework the implementation in that direction if the maintainers prefer it.

For now, this PR introduces `AwaitPrompt` as a distinct command to make the behaviour explicit and avoid changing existing `Wait` semantics.

## How it works

A central challenge discussed in #70 is that ttyd has no API to expose the state of child processes (confirmed by the ttyd maintainer in [tsl0922/ttyd#1009](https://github.com/tsl0922/ttyd/issues/1009#issuecomment-1295718599)). VHS communicates with ttyd through a browser – keystrokes go in, rendered canvas frames come out. There is no side channel for process lifecycle information.

This implementation sidesteps that limitation entirely. Rather than querying ttyd about process state, it reads the raw terminal output stream that already flows from ttyd through xterm.js. The approach follows the direction suggested by @normanr in #70 (using shell prompt detection via escape sequences) and is similar to what @kotborealis implemented with `PROMPT_COMMAND` in [cassette_deck](https://github.com/kotborealis/cassette_deck/blob/develop/src/x/yaftx.c#L72).

The mechanism:

1. An invisible [FinalTerm shell integration sequence](https://invisible-island.net/xterm/ctlseqs/ctlseqs.html) (`ESC ] 133;A BEL` – "prompt start") is embedded in each shell's prompt string. OSC 133 is a recognised standard supported by iTerm2, WezTerm, Kitty, Ghostty, VS Code terminal, and others.
2. When a command finishes and the shell renders a new prompt, this marker flows through: **shell → ttyd → websocket → xterm.js → `term.write()`**
3. A JavaScript hook on `term.write()` scans for `\x1b]133;A` in the raw data *before* xterm.js processes it, incrementing a counter each time. The match is on the prefix only, so both BEL- and ST-terminated variants (cmd.exe) are detected.
4. `AwaitPrompt` records the current counter and polls until it increases (or the timeout expires)

Because the marker is emitted by the shell's own prompt rendering, it fires precisely when the shell is ready for the next command – regardless of how long the previous command took.

### Cross-platform shell support

Every shell supported by VHS embeds the OSC 133;A marker in its prompt:

| Shell | Marker mechanism | Notes |
|-------|-----------------|-------|
| **bash** | `\e]133;A\a` in `PS1` | Uses `\[...\]` to mark as non-printing |
| **zsh** | `\e]133;A\a` in `PROMPT` | Uses `%{...%}` to mark as non-printing |
| **fish** | `printf '\e]133;A\a'` in `fish_prompt` | Emitted before the visible prompt |
| **PowerShell** | `[Console]::Write(ESC + ']133;A' + BEL)` | Direct console write in `prompt` function |
| **pwsh** | Same as PowerShell | Same mechanism |
| **cmd.exe** | `$E]133;A$E\` in `prompt` | Uses ST (`ESC \`) instead of BEL, since cmd.exe's `prompt` command has no BEL code |
| **nushell** | `print -n` with marker in `PROMPT_COMMAND` | Raw bytes embedded by Go; nushell passes them through verbatim |
| **osh** | Same as bash | Shares bash's `PS1` format |
| **xonsh** | Raw marker bytes in `PROMPT` | Raw bytes embedded by Go via `-D` flag |

A test (`TestShellPromptMarker`) verifies that all 9 shells include the `133` marker. Both BEL and ST are valid OSC terminators per the spec and are accepted by all major terminal emulators.

The remaining FinalTerm markers – `B` (prompt end), `C` (command start), `D` (command finished with exit code) – are natural follow-ups if richer shell integration is needed later.

## Motivation / Related discussions

- **#70** – "Wait for command to finish before continuing" – the original discussion. @normanr proposed shell prompt detection via escape sequences; @kotborealis demonstrated a working `PROMPT_COMMAND` approach
- **#537** – "How to auto-advance on completion" – `Wait` partially addressed this but requires knowing the expected output pattern
- **#151**, **#166** – duplicates of #70 showing continued demand
- **#481** – "VHS doesn't block on command completion"
- **#611** – "auto-delays between commands"
- **#691** – `PROMPT_COMMAND` leaking into VHS recordings

## Test plan

- [x] Parser tests: bare `AwaitPrompt`, with `@timeout` (seconds, milliseconds, minutes)
- [x] Command count test updated for the new command
- [x] Shell prompt marker test: all 9 shells verified to include OSC 133;A
- [x] Full test suite passes (`go test ./...`)
- [ ] Manual testing with real tape files across available shells

---

> [!NOTE]
> This PR was authored with AI assistance (Claude Code).

### DIFF
--- a/command.go
+++ b/command.go
@@ -210,7 +210,7 @@ func ExecuteWait(c parser.Command, v *VHS) error {
 }
 
 // ExecuteAwaitPrompt waits for the shell to emit a new prompt marker.
-// It detects prompt markers (OSC 7777) that are embedded in each shell's prompt
+// It detects prompt markers (OSC 133;A) that are embedded in each shell's prompt
 // configuration. Unlike Wait (which matches terminal content), AwaitPrompt
 // detects when the shell has finished executing a command and is ready for input.
 func ExecuteAwaitPrompt(c parser.Command, v *VHS) error {
--- a/command.go
+++ b/command.go
@@ -39,37 +39,38 @@ type CommandFunc func(c parser.Command, v *VHS) error
 
 // CommandFuncs maps command types to their executable functions.
 var CommandFuncs = map[parser.CommandType]CommandFunc{
-	token.BACKSPACE:   ExecuteKey(input.Backspace),
-	token.DELETE:      ExecuteKey(input.Delete),
-	token.INSERT:      ExecuteKey(input.Insert),
-	token.DOWN:        ExecuteKey(input.ArrowDown),
-	token.ENTER:       ExecuteKey(input.Enter),
-	token.LEFT:        ExecuteKey(input.ArrowLeft),
-	token.RIGHT:       ExecuteKey(input.ArrowRight),
-	token.SPACE:       ExecuteKey(input.Space),
-	token.UP:          ExecuteKey(input.ArrowUp),
-	token.TAB:         ExecuteKey(input.Tab),
-	token.ESCAPE:      ExecuteKey(input.Escape),
-	token.PAGE_UP:     ExecuteKey(input.PageUp),
-	token.PAGE_DOWN:   ExecuteKey(input.PageDown),
-	token.SCROLL_UP:   ExecuteScroll(-1),
-	token.SCROLL_DOWN: ExecuteScroll(1),
-	token.HIDE:        ExecuteHide,
-	token.REQUIRE:     ExecuteRequire,
-	token.SHOW:        ExecuteShow,
-	token.SET:         ExecuteSet,
-	token.OUTPUT:      ExecuteOutput,
-	token.SLEEP:       ExecuteSleep,
-	token.TYPE:        ExecuteType,
-	token.CTRL:        ExecuteCtrl,
-	token.ALT:         ExecuteAlt,
-	token.SHIFT:       ExecuteShift,
-	token.ILLEGAL:     ExecuteNoop,
-	token.SCREENSHOT:  ExecuteScreenshot,
-	token.COPY:        ExecuteCopy,
-	token.PASTE:       ExecutePaste,
-	token.ENV:         ExecuteEnv,
-	token.WAIT:        ExecuteWait,
+	token.BACKSPACE:     ExecuteKey(input.Backspace),
+	token.DELETE:        ExecuteKey(input.Delete),
+	token.INSERT:        ExecuteKey(input.Insert),
+	token.DOWN:          ExecuteKey(input.ArrowDown),
+	token.ENTER:         ExecuteKey(input.Enter),
+	token.LEFT:          ExecuteKey(input.ArrowLeft),
+	token.RIGHT:         ExecuteKey(input.ArrowRight),
+	token.SPACE:         ExecuteKey(input.Space),
+	token.UP:            ExecuteKey(input.ArrowUp),
+	token.TAB:           ExecuteKey(input.Tab),
+	token.ESCAPE:        ExecuteKey(input.Escape),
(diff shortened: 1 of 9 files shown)


aymanbagabas (commented):
Thank you, @MattieTK, for your work here. I wonder if instead of using our own OSC7777 sequences, we could use FinalTerm shell integration sequences OSC133 to indicate prompt begin, end, command begin, and command end 🤔

aymanbagabas:
> Great idea, I've swapped the custom string for 133;A which is 'new prompt' and practically the same thing as far as we're concerned in this case.

I don't see any new commits. Did you push your changes? 😄

MattieTK:
> > Great idea, I've swapped the custom string for 133;A which is 'new prompt' and practically the same thing as far as we're concerned in this case.
> 
> I don't see any new commits. Did you push your changes? 😄

Sorry @aymanbagabas! I of course left this locally and then promptly went on holiday. 

Merged now and rebased.

MattieTK:
Hey @andreynering - sorry to ping, any chance you could give this a look soon? I'd love to get more shell recordings from vhs in our ouput!

command.go
```diff
@@ -208,6 +209,47 @@ func ExecuteWait(c parser.Command, v *VHS) error {
...
+	timeout := v.Options.WaitTimeout
+	if c.Options != "" {
+		t, err := time.ParseDuration(c.Options)
+		if err != nil {
+			return fmt.Errorf("failed to parse duration: %w", err)
+		}
+		timeout = t
+	}
+
+	// Record the current prompt count so we can detect the next one.
```
andrinoff: The baseline is read when `AwaitPrompt` starts, so if the previous command already finished and printed its prompt before this runs (common for fast commands), the new prompt is counted into the baseline and this will sit until timeout. Consider grabbing the count before `Enter` is pressed, or resetting the counter when a command is typed

Issue #458, opened 2024-03-21
Escaped prompt when running VHS in a nix shell


meowgorithm:
Example below.

<img width="500" src="https://github.com/charmbracelet/vhs/assets/25087/504b9dda-2ad5-4384-aecc-557611e16352">

pspiagicw:
The same thing happens when you run `vhs record > some.tape`. The entire prompt is having escape characters.

I found this discussion [here](https://github.com/NixOS/nix/issues/730),  which references this issue
.I don't have much information on why that happens.

But to solve it add `bashInteractive` to the package lists of whichever system you use Nix Flakes `configuration.nix` or Home Manager.


I use `flake.nix` to manage dependencies per project and adding it to the package lists solved the above issues for me.

PR #709, opened 2026-02-20
feat: add PromptColor setting for customisable shell prompt colour

MattieTK:
## Summary

Adds a `Set PromptColor` setting that lets users change the colour of VHS's default prompt across all supported shells.

### Syntax

```tape
Set PromptColor "#FF8000"    # orange prompt
Set PromptColor "#00FF00"    # green prompt
```

The default remains `#5B56E0` (the existing blue-purple), so existing tapes are unaffected.

## Motivation

Prompt customisation is one of the most requested features in VHS. Issue #419 ("How can I change Shell Prompt") has been open since mid-2023 and has driven users to alternative tools:

> *"I've switched to asciinema at the moment just because of this issue"* – @ocervell
>
> *"This issue is basically forcing my hand to use asciinema"* – @artob
>
> *"Setting PS1 with the Env command has no effect on the prompt for bash"* – @artob

There was a previous attempt in PR #469 where @Delta456 tried reading `os.Getenv("PROMPT")` at module init time, but the environment wasn't set yet at that point. @maaslalani removed the hack and noted: "Let's remove this hack for now and keep thinking how to solve PROMPTs."

The current workaround suggested in #419 involves a `Hide`/`Type`/`Enter`/`clear`/`Show` sequence, which is fragile and leaves artefacts.

### The colour clash problem

VHS already lets users control themes, margins, and window bars – but the prompt colour is hardcoded to `#5B56E0` and cannot be adjusted. This creates two problems:

**Accessibility:** Themes with blue or purple backgrounds make the prompt nearly invisible. The `>` disappears against backgrounds of similar hue and luminance:

| `Set Theme "Ocean"` | `Set Theme "Fairyfloss"` |
|:---:|:---:|
| Ocean theme – prompt invisible on blue background | Fairyfloss theme – prompt lost in purple-grey background |

**Aesthetic clash:** When users style their recordings with coloured margins or borders, the fixed purple prompt conflicts with the chosen palette:

| `Set MarginFill "#FF6B2B"` |
|:---:|
| Orange margin with clashing purple prompt |

### With `Set PromptColor`

Adding a single line to the tape fixes both problems:

| `Set Theme "Ocean"` | `Set Theme "Fairyfloss"` |
|:---:|:---:|
| `Set PromptColor "#00FF88"` | `Set PromptColor "#FF857A"` |
| Ocean with green prompt | Fairyfloss with pink prompt |

| `Set MarginFill "#FF6B2B"` |
|:---:|
| `Set PromptColor "#FF6B2B"` |
| Orange margin with matching orange prompt |

### Why colour first, not the full prompt?

This PR deliberately limits scope to the prompt *colour* rather than the full prompt string (symbol, format, etc.). Colour is the simpler case – it doesn't affect prompt width or terminal layout, and the `>` symbol is consistent across shells. Full prompt symbol customisation is demonstrated in #710 (`Set Prompt`), which builds directly on the `ShellConfig()` refactoring introduced here.

## How it works

1. **`ShellConfig(name, promptColor)`** replaces the old static `Shells` map entries. Instead of hardcoded `Shell{Env: ..., Command: ...}` structs, each shell's configuration is generated dynamically with the chosen colour embedded in the appropriate format (ANSI RGB for bash/osh/nushell/xonsh, hex for zsh/fish, `System.Drawing.Color` for PowerShell).

2. **`hexToRGB()`** parses hex colour strings (with or without `#`) to RGB components for shells that need numeric values.

3. **Evaluator integration**: `Set PromptColor` is processed in the same early pass as `Set Shell`, before the tty is started. This was the key insight that #469 missed – the colour needs to be known *before* `buildTtyCmd` runs, so it's handled alongside Shell in the evaluator's pre-start loop.

### Shell support

| Shell | Colour format | Notes |
|-------|--------------|-------|
| **bash** | ANSI `38;2;R;G;B` in PS1 | RGB values from hex |
| **zsh** | `%F{#RRGGBB}` in PROMPT | Hex without `#` prefix |
| **fish** | `set_color RRGGBB` in fish_prompt | Hex without `#` prefix |
| **PowerShell** | `System.Drawing.Color::FromArgb(R,G,B)` | RGB values |
| **pwsh** | Same as PowerShell | Same mechanism |
| **cmd.exe** | Not supported | cmd.exe's `prompt` command has no colour codes; keeps default |
| **nushell** | ANSI `38;2;R;G;B` in PROMPT_COMMAND | RGB values |
| **osh** | Same as bash | Shares bash's PS1 format |
| **xonsh** | ANSI `38;2;R;G;B` in PROMPT | RGB values |

## Follow-up: Set Prompt (#710)

The `ShellConfig()` refactoring here also enables customising the prompt *symbol*. #710 demonstrates this by adding a `Set Prompt` setting that accepts any string:

```tape
Set PromptColor "#FF8000"
Set Prompt "$"    # bash-style
```

That PR adds a single parameter to `ShellConfig()` – a direct payoff of the structural work in this PR.

## Related discussions

- **#419** – "How can I change Shell Prompt" – the primary motivation. Open since 2023, multiple users affected
- **#469** – PR that attempted prompt customisation via `Env PROMPT`, removed as hacky
- **#458** – Escaped prompt in nix shell – related prompt rendering issue
- **#39** – Strange prompt output at start of gif – prompt configuration issues on macOS

## Test plan

- [x] `TestHexToRGB`: hex parsing with `#`, without `#`, invalid length, empty string
- [x] `TestShellConfigReturnsNonNil`: all 9 shells return valid configs
- [x] `TestShellConfigCustomColor`: custom colour (#FF8000) correctly embedded in bash RGB and zsh hex
- [x] `TestShellConfigDefaultColor`: default colour produces original RGB values
- [x] `TestShellConfigUnknownShell`: unknown shell returns nil
- [x] Full test suite passes (`go test ./...`)
- [ ] Manual testing with real tape files

---

> [!NOTE]
> This PR was authored with AI assistance (Claude Code).


MattieTK:
Hey @meowgorithm - sorry to ping, any chance you could give this a look soon? I'd love to get more shell recordings from vhs in our ouput!

maaslalani:
I like this and I think this is reasonable to merge. I do wonder though if we should just use a theme color and allow this to be set in the theme instead?

MattieTK:
@maaslalani happy to change it to that, except I think that would generally lock it to themes, which may be overkill for what you want: see the last example where really I just want the already supported `MarginFill` setting, and to change the `PromptColor`. I would argue that if we did want everything to be in the theme then you wouldn't already have theme-adjacent options like `MarginFill` (although admittedly that sits alone at the moment)

maaslalani:
@MattieTK, very good point. Let's ship it 🚢

maaslalani:
I think ideally though every color should be set by the theme.

maaslalani (approved):

meowgorithm:
@MattieTK: mind fixing the build issues?

MattieTK:
> @MattieTK: mind fixing the build issues?

Claude has had a go if you could run the actions

MattieTK:
@meowgorithm the lint failures are sorted and `main` is merged in. 

Workflow runs need maintainer approval for a first-time contributor. Could you kick it off when you get a moment?

(Edited massive LLM response)

command.go
```diff
@@ -751,6 +756,12 @@ func ExecuteSetCursorBlink(c parser.Command, v *VHS) error {
 	return nil
 }
 
+// ExecuteSetPromptColor sets the prompt color.
+func ExecuteSetPromptColor(c parser.Command, v *VHS) error {
+	v.Options.PromptColor = c.Args
```
andrinoff, line 761: Any string is accepted here with no validation, so a typo like `Set PromptColor #5B56E` silently gives a black prompt on bash instead of an error. Worth checking it's a valid 6-digit hex and returning an error otherwise, like the other settings do for bad input

shell.go
```diff
@@ -13,64 +18,97 @@ const (
...
 			"-NoLogo",
 			"-NoExit",
 			"-NoProfile",
 			"-Command",
-			`Set-PSReadLineOption -HistorySaveStyle SaveNothing; function prompt { Write-Host '>' -NoNewLine -ForegroundColor Blue; return ' ' }`,
-		},
-	},
-	pwsh: {
-		Command: []string{
+			fmt.Sprintf(`Set-PSReadLineOption -HistorySaveStyle SaveNothing; function prompt { Write-Host '>' -NoNewLine -ForegroundColor ([System.Drawing.Color]::FromArgb(%d,%d,%d)); return ' ' }`, r, g, b),
```
andrinoff, line 64: `Write-Host -ForegroundColor` only accepts named console colours (Black, Blue, etc.), not a `System.Drawing.Color`, so this will throw a "cannot convert" error and likely break the prompt on Windows PowerShell entirely. Using an ANSI escape sequence in the prompt string (as bash/nushell do) would be safer.

shell.go
```diff
@@ -13,64 +18,97 @@ const (
...
-		Command: []string{"nu", "--execute", "$env.PROMPT_COMMAND = {'\033[;38;2;91;86;224m>\033[m '}; $env.PROMPT_COMMAND_RIGHT = {''}"},
-	},
-	osh: {
-		Env:     []string{"PS1=\\[\\e[38;2;90;86;224m\\]> \\[\\e[0m\\]"},
-		Command: []string{"osh", "--norc"},
-	},
-	xonsh: {
-		Command: []string{"xonsh", "--no-rc", "-D", "PROMPT=\033[;38;2;91;86;224m>\033[m "},
-	},
+			fmt.Sprintf(`Set-PSReadLineOption -HistorySaveStyle SaveNothing; Function prompt { Write-Host -ForegroundColor ([System.Drawing.Color]::FromArgb(%d,%d,%d)) -NoNewLine '>'; return ' ' }`, r, g, b),
```
andrinoff, line 74: Same problem as above for `pwsh`, plus `System.Drawing` isn't guaranteed to be loaded on Linux/macOS

shell.go
```diff
@@ -13,64 +18,97 @@ const (
...
+	switch name {
+	case bash:
+		return []string{
+				fmt.Sprintf("PS1=\\[\\e[38;2;%d;%d;%dm\\]> \\[\\e[0m\\]", r, g, b),
+				"BASH_SILENCE_DEPRECATION_WARNING=1",
+			},
+			[]string{"bash", "--noprofile", "--norc", "--login", "+o", "history"}
+	case zsh:
+		return []string{fmt.Sprintf(`PROMPT=%%F{#%s}> %%F{reset_color}`, hexNoHash)},
+			[]string{"zsh", "--histnostore", "--no-rcs"}
```
andrinoff, line 47: The hex is passed straight through to `zsh` here, while bash uses the parsed RGB. If the input is something invalid, bash falls back to black but zsh gets garbage in `%F{}`, so the same bad input behaves differently per shell.

This is going back to the first comment, we should validate

shell.go
```diff
@@ -13,64 +18,97 @@ const (
...
-		Env:     []string{"PS1=\\[\\e[38;2;90;86;224m\\]> \\[\\e[0m\\]"},
-		Command: []string{"osh", "--norc"},
-	},
-	xonsh: {
-		Command: []string{"xonsh", "--no-rc", "-D", "PROMPT=\033[;38;2;91;86;224m>\033[m "},
-	},
+			fmt.Sprintf(`Set-PSReadLineOption -HistorySaveStyle SaveNothing; Function prompt { Write-Host -ForegroundColor ([System.Drawing.Color]::FromArgb(%d,%d,%d)) -NoNewLine '>'; return ' ' }`, r, g, b),
+		}
+	case cmdexe:
+		return nil, []string{"cmd.exe", "/k", "prompt=^> "}
```
andrinoff, line 77: cmd.exe silently ignores the colour setting. That is probably fine, but we should note this in a comment or docs

andrinoff (changes requested):
Thank you for your contribution. Please, note the comments below. Also, adding this new command to the docs/ README would be nice.

PR #706, opened 2026-02-19
Feat/asciicinema

taigrr:
- [X] I have read [`CONTRIBUTING.md`](https://github.com/charmbracelet/.github/blob/main/CONTRIBUTING.md).
- [X] I have created a discussion that was approved by a maintainer (for new features).

Closes #225

PR #711, opened 2026-02-23
feat: add asciicast v2 output format

### LINKED ISSUE #109: Asciicast Output
This is probably out of scope but would be nice to have [asciicast](https://asciinema.org/) output (`Output: out.cast`).
I initally wanted `gif` for the readme and `asciicast` for the mdbook documentation but then decided against `gif` due to it's size.
Managed to record it by simply spawning an `asciinema` recording within `tmux`.
Keeping size in sync with the other outputs could be challenge but as there are also `gif`, `svg`,.. generators available for `asciicast`.
It might even provide an alternative to the chrome dependency (which pushes my cpu to the limit).

- width/height with [tmux](https://github.com/tmux/tmux)
- hide/show support with [pause/resume](https://github.com/asciinema/asciinema/pull/340)

```elixir
Hide
Type "tmux"
Enter
Type "tmux resize-window -x 108 -y 24"
Enter
Type "asciinema rec --overwrite out.cast"
Enter
Sleep 1s
Show

# ...

Hide
Type "exit"
Enter
Show
```

## Example

Source:  https://github.com/rsteube/carapace-bin/blob/18ca32ea6093b965e4949a13ccce85f9108ca65c/carapace-bin.tape

### Asciicast

[asciicast](https://asciinema.org/a/533283)

### Gif

GIF output with [agg](https://github.com/asciinema/agg) (ignore the missing symbols, didn't get the font right)

```sh
agg --font-family "RobotoMono Nerd Font Mono,Noto Color Emoji"  carapace-bin.cast out.gif
```

out

### Svg

SVG output with [svg-term-cli](https://github.com/marionebl/svg-term-cli) (nice output but high load for playing it)

```sh
cat carapace-bin.cast | svg-term --out out.svg
```

out

related #105

veeceey:
Adds support for generating `.cast` (asciicast v2) files as an output format, so you can do:

```
Output demo.cast
```

This has been a requested feature since the early days (#109) - being able to export recordings as asciicast means you can use them with asciinema players, convert to SVG with svg-term-cli, or generate GIFs with agg.

### How it works

- Added `Asciicast` field to `VideoOutputs` so `.cast` is a first-class output alongside GIF/MP4/WebM
- During recording, terminal buffer snapshots are captured each frame (only when `.cast` output is requested to avoid overhead)
- At render time, `GenerateAsciicast` writes the asciicast v2 JSON Lines format: a header line with version/width/height/timestamp, followed by `"o"` (output) events with deduplicated content

### Testing

```
$ go test ./...
ok      github.com/charmbracelet/vhs    4.806s
ok      github.com/charmbracelet/vhs/lexer      5.044s
ok      github.com/charmbracelet/vhs/parser      5.272s
ok      github.com/charmbracelet/vhs/token      5.482s
```

Closes #109

### DIFF
--- a/asciicast.go
+++ b/asciicast.go
@@ -1,0 +1,149 @@
+package main
+
+import (
+	"encoding/json"
+	"fmt"
+	"log"
+	"os"
+	"path/filepath"
+	"strings"
+	"time"
+)
+
+// AsciicastHeader is the header of an asciicast v2 file.
+type AsciicastHeader struct {
+	Version   int               `json:"version"`
+	Width     int               `json:"width"`
+	Height    int               `json:"height"`
+	Timestamp int64             `json:"timestamp,omitempty"`
+	Env       map[string]string `json:"env,omitempty"`
+}
+
+// AsciicastEvent represents a single event in an asciicast v2 recording.
+// Format: [time, event-type, event-data]
+type AsciicastEvent struct {
+	Time float64
+	Type string
+	Data string
+}
+
+// MarshalJSON implements custom JSON marshaling for AsciicastEvent
+// to produce the [time, type, data] array format.
+func (e AsciicastEvent) MarshalJSON() ([]byte, error) {
+	return json.Marshal([]interface{}{e.Time, e.Type, e.Data})
+}
+
+const cast = ".cast"
+
+// GenerateAsciicast creates an asciicast v2 file from terminal buffer snapshots
+// captured during recording. It writes the asciicast v2 format (JSON Lines)
+// with a header line followed by event lines.
+func GenerateAsciicast(v *VHS) error {
+	path := v.Options.Video.Output.Asciicast
+	if path == "" {
+		return nil
+	}
+
+	log.Println(GrayStyle.Render("Creating " + path + "..."))
+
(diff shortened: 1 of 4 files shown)


veeceey:
Good question! I went with v2 for a practical reason: it has the widest ecosystem support right now. Most of the popular tools that consume asciicast files — asciinema-player, svg-term, agg, and others — have mature, battle-tested v2 support. The v3 format was introduced with asciinema 3.0 and while it does bring improvements (like the header changes and new event types), tooling support for v3 is still catching up. If I had implemented v3, the output files would not be usable with the majority of existing tools people are already relying on.

That said, adding v3 support down the line would be a natural follow-up once the broader ecosystem has adopted it more widely. The internal structure here is straightforward enough that supporting v3 later should be pretty painless.

drupol:
There's already a PR about this in https://github.com/charmbracelet/vhs/pull/706, it would be nice to collaborate to avoid potential frustrations.

veeceey:
Good point @drupol, I wasn't aware of #706 when I opened this. Happy to collaborate -- I'll take a look at that PR and see if there's anything from my implementation that could complement it, or if we should just consolidate there.

veeceey:
looks like #706 already covers asciicast output and is further along - closing this in favor of that one.

PR #726, opened 2026-03-22
feat: add Subtitle command for text overlays on recordings

gtmax:
Discussion: #727

## Summary

Adds a `Subtitle` command that renders text overlays on terminal recordings — useful for demos, tutorials, and walkthroughs where you want to explain what's happening on screen.

Related: #173 (Progress bars & Overlays)
Related: #164 (display key names — the overlay infrastructure added here could be extended to support keystroke display)

## Tape Syntax

```tape
# Show a subtitle
Subtitle "This is what's happening"

# Change the subtitle
Subtitle "Now something else"

# Hide the subtitle
Subtitle ""
```

## Settings

```tape
Set SubtitleFontSize 28
Set SubtitleFontFamily "system-ui"
Set SubtitleColor "#ffffff"
Set SubtitleBackground "rgba(0,0,0,0.75)"
Set SubtitlePosition "bottom"       # top, center, bottom
Set SubtitlePadding 12
Set SubtitleBorderRadius 8
```

The subtitle DSL is deliberately kept simple for now — settings cover the common styling needs, and the implementation can be swapped or extended later without changing the tape syntax.

## Example

Here's a real-world demo GIF using subtitles to narrate a terminal workflow ([wotr](https://github.com/gtmax/wotr)):

wotr demo with subtitles

## Implementation: Canvas Overlay Stream

The subtitle is rendered onto a separate `<canvas>` element and captured as a third frame stream alongside the existing text and cursor streams. ffmpeg composites it on top of the fully styled terminal frame (after padding, window bar, border radius, and margins).

### Why Canvas and not HTML/CSS?

HTML/CSS would give richer styling, but `Element.Screenshot()` and `Page.Screenshot()` in Chrome DevTools Protocol do not preserve transparency — they always composite against the page background. Since VHS needs transparent PNGs for its ffmpeg compositing pipeline, only `Canvas.toDataURL()` (via `CanvasToImage`) works, as it returns canvas content with alpha channel intact.

### Why not ASS subtitles (burned in via ffmpeg)?

ffmpeg's `ass`/`subtitles` filter requires `libass`, which is not included in standard Homebrew ffmpeg builds. Requiring users to compile ffmpeg with `--enable-libass` would break the install experience. The canvas approach requires no additional ffmpeg dependencies beyond what VHS already uses.

### Performance

When no `Subtitle` commands are present in the tape, the overlay stream is not created and there is zero overhead. The tape is pre-scanned for `Subtitle` commands before recording begins.

## Changes

- `token/token.go` — Add `SUBTITLE` command token and subtitle setting tokens
- `parser/parser.go` — Add `parseSubtitle()` parser and register in command dispatcher
- `command.go` — Add `ExecuteSubtitle` and subtitle setting executors
- `vhs.go` — Add `SubtitleOptions`, overlay canvas setup, `renderOverlay()` method
- `evaluator.go` — Pre-scan for `Subtitle` commands to enable overlay stream
- `video.go` — Add overlay frame format, third input stream, `HasOverlay` flag
- `ffmpeg.go` — Add `WithOverlay` filter builder, `overlayStream` on `StreamBuilder`
- `command_test.go` — Update command count
- `parser/parser_test.go` — Add subtitle parser tests


gtmax:
Per the contributing guidelines, opened a Discussion for this feature: #727

Converting to draft until the team has a chance to review the approach.

PR #737, opened 2026-03-31
fix: kill ttyd process on early exit from Evaluate

### LINKED ISSUE #738: ttyd process leaked on early exit from Evaluate
## Description

When `Evaluate()` returns early — between `Start()` and the `Record()`/`teardown()` path — the `ttyd` process is left running indefinitely.

The deferred cleanup at `evaluator.go:45` only calls `v.close()`, which is set to `vhs.browser.Close`. The `terminate()` method that kills both browser **and** ttyd is only reachable via `teardown()`, which requires `Record()` to have started.

## Affected code paths

Any early return after `Start()` succeeds but before `Record()` begins:

- `Page.Wait` fails (`evaluator.go:50-53`)
- A `SET` command fails (`evaluator.go:59-63`)
- Video dimensions too small (`evaluator.go:78-91`)
- A `Hide` block command fails (`evaluator.go:106-109`)

Additionally, `Start()` itself doesn't clean up ttyd if browser launch or page creation fails after ttyd is already started.

## Reproduction

```bash
# Create a tape that triggers an early SET error
echo 'Set Width 0' > /tmp/bad.tape
echo 'Type "hello"' >> /tmp/bad.tape

# Run vhs (it will error due to bad dimensions)
vhs /tmp/bad.tape

# Check for orphaned ttyd
ps aux | grep ttyd
# ttyd process is still running
```

## Expected behavior

All `ttyd` processes spawned by VHS should be killed when VHS exits, regardless of which code path is taken.

tomenden:
Fixes #738

## Summary

`ttyd` processes are left running after VHS exits when `Evaluate()` returns early (before `Record()`/`teardown()` are reached).

The deferred cleanup at `evaluator.go:45` only calls `v.close()`, which is set to `vhs.browser.Close` — it never kills the `ttyd` process. The `terminate()` method that properly kills both browser and ttyd is only called via `teardown()`, which is only reachable after `Record()` starts.

**Affected code paths** (all between `Start()` and `Record()`):
- `Page.Wait` fails (line 50-53)
- A `SET` command fails (line 59-63)
- Video dimensions too small (line 78-91)
- A `Hide` block command fails (line 106-109)

Additionally, `Start()` itself didn't clean up ttyd if browser launch or page creation failed after ttyd was already started.

## Fix

- **`evaluator.go`**: The deferred cleanup now also kills the ttyd process, so every early-return path is covered.
- **`vhs.go`**: `Start()` now kills ttyd if subsequent browser/page setup fails, preventing orphaned processes even on initialization errors.

## Test plan

- [x] `go build ./...` passes
- [x] Run `vhs` with a tape that triggers an early exit (e.g., invalid SET command), then verify no orphaned `ttyd` processes remain (`ps aux | grep ttyd`)
- [x] Run `vhs` normally and verify recordings still work correctly

### DIFF
--- a/evaluator.go
+++ b/evaluator.go
@@ -42,7 +42,12 @@ func Evaluate(ctx context.Context, tape string, out io.Writer, opts ...Evaluator
 	if err := v.Start(); err != nil {
 		return []error{err}
 	}
-	defer func() { _ = v.close() }()
+	defer func() {
+		_ = v.close()
+		if v.tty != nil && v.tty.Process != nil {
+			_ = v.tty.Process.Kill()
+		}
+	}()
 
 	// Let's wait until we can access the window.term variable.
 	//
--- a/vhs.go
+++ b/vhs.go
@@ -140,11 +140,14 @@ func (vhs *VHS) Start() error {
 	enableNoSandbox := os.Getenv("VHS_NO_SANDBOX") != ""
 	u, err := launcher.New().Leakless(false).Bin(path).NoSandbox(enableNoSandbox).Launch()
 	if err != nil {
+		_ = vhs.tty.Process.Kill()
 		return fmt.Errorf("could not launch browser: %w", err)
 	}
 	browser := rod.New().ControlURL(u).MustConnect()
 	page, err := browser.Page(proto.TargetCreateTarget{URL: fmt.Sprintf("http://localhost:%d", port)})
 	if err != nil {
+		browser.MustClose()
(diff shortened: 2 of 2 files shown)


tomenden:
Fixes #738

tomenden:
LMK if uninterested in this and I will close 🙏 
Thanks for an amazing project!

meowgorithm (approved):

PR #763, opened 2026-07-10
feat: rows/cols for axises

DaltonSW:
Adds `Set Rows`/`Set Columns` commands to size the terminal in character
cells instead of pixels. The pixel size is derived by probing the live
xterm.js terminal's own fit/measurement APIs and iterating to an exact
match, so it stays correct across font/theme changes without depending
on any of xterm.js's private internals.

- [x] I have created a discussion that was approved by a maintainer (for new features). (I got Discord approval, if that counts)

### DIFF
--- a/vhs.go
+++ b/vhs.go
@@ -21,18 +21,20 @@ import (
 
 // VHS is the object that controls the setup.
 type VHS struct {
-	Options      *Options
-	Errors       []error
-	Page         *rod.Page
-	browser      *rod.Browser
-	TextCanvas   *rod.Element
-	CursorCanvas *rod.Element
-	mutex        *sync.Mutex
-	started      bool
-	recording    bool
-	tty          *exec.Cmd
-	totalFrames  int
-	close        func() error
+	Options        *Options
+	Errors         []error
+	Page           *rod.Page
+	browser        *rod.Browser
+	TextCanvas     *rod.Element
+	CursorCanvas   *rod.Element
+	mutex          *sync.Mutex
+	started        bool
+	recording      bool
+	tty            *exec.Cmd
+	totalFrames    int
+	close          func() error
+	widthExplicit  bool
+	heightExplicit bool
 }
 
 // Options is the set of options for the setup.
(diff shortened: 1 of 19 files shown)


meowgorithm:
@DaltonSW mind fixing the build errors?

DaltonSW:
@meowgorithm Done! I think I misunderstood LFS stuff. :sweat_smile: 

Did you want me to clean up any of the `govulncheck` or `golint` stuff?

meowgorithm:
Apologies for the slow reply @DaltonSW — yes, please. We'll need CI green in order to merge this.

vhs.go
```diff
@@ -158,38 +160,135 @@ func (vhs *VHS) Start() error {
...
+			(style.Columns > 0 && cols < style.Columns) ||
+			(style.Rows > 0 && rows < style.Rows)
+		if !tooSmall {
+			cellWidth = float64(probeWidth) / float64(cols)
+			cellHeight = float64(probeHeight) / float64(rows)
+			break
+		}
+		probeWidth = double(probeWidth)
+		probeHeight = double(probeHeight)
+	}
```
andrinoff, line 265: silent failure. if `!tooSmall` is never satisfied `cellWidth` and `cellHeight` stay 0

vhs.go
```diff
@@ -158,38 +160,135 @@ func (vhs *VHS) Start() error {
...
+
+	contentWidth := probeWidth
+	contentHeight := probeHeight
+	if style.Columns > 0 {
+		contentWidth = int(math.Round(cellWidth * float64(style.Columns)))
+	}
+	if style.Rows > 0 {
+		contentHeight = int(math.Round(cellHeight * float64(style.Rows)))
+	}
```
andrinoff, line 284: it will result in an 0x0 image, reasons in comment 1

vhs.go
```diff
@@ -158,38 +160,135 @@ func (vhs *VHS) Start() error {
...
+			converged = false
+		}
+		if style.Rows > 0 && rows != style.Rows {
+			contentHeight += int(math.Round(float64(style.Rows-rows) * cellHeight))
+			converged = false
+		}
+		if converged {
+			break
+		}
+	}
```
andrinoff, line 307: this will give up **silently** after 5 attempts

andrinoff:
Hey @DaltonSW, thank you for this PR. Regarding the token variables, I talked to Christian, lets add an exception for this rule (in this repo)

andrinoff (changes requested):
Thank you, @DaltonSW! Just a few things

DaltonSW:
Thanks for the review @andrinoff — good catches, all three were the same silent-failure problem. Fixed in e45e4b0.

`resolveRowsColumns` now returns an error, and `Setup` returns one too so it surfaces through the existing `Evaluate` error path rather than being swallowed:

- **Probe loop never succeeds** — the loop now tracks whether it ever measured the terminal and returns early, so the zero `cellWidth`/`cellHeight` can no longer reach the `contentWidth`/`contentHeight` computation. That covers the 0x0 image case too.
- **Correction loop exhausts its attempts** — it no longer falls through; it reports what was requested vs. the closest grid it reached.

Both errors name the requested grid and the last measurement, e.g.:

```
could not fit 400 columns x 200 rows: terminal measured 566 columns x 131 rows at 69120x30720 pixels after 6 attempts
```

Build, vet, tests, and `golangci-lint` are all clean.

andrinoff (changes requested):
We should round up to even numbers, because MP4 and WebM Video dont like odd numbers as the resolution

andrinoff (approved):

PR #771, opened 2026-07-31
fix: honor Set Framerate on overlay input streams

### LINKED ISSUE #750: `Set Framerate` doesnt work :(
**Describe the bug**
When exporting a tape to `.mp4` and using the `Set Framerate` command (example `Set Framerate 50`), the resulting video is still encoded at 25 fps. This happens regardless of the framerate set in the tape file. 

I dug into the source and it seems to be caused by two things in the ffmpeg pipeline:
1. If you use a `MarginFill` color, `WithMargin()` creates the background using ffmpeg's `lavfi` color filter (`color=%s:s=%dx%d`). Since no `rate` is passed, lavfi defaults to 25 fps. Because this 25fps stream is used as the base for the `overlay` filter, ffmpeg forces the entire output video to match that 25fps.
2. Even without a margin, the final ffmpeg command built in `video.go` is missing an explicit `-r <framerate>` flag right before the output file. Without it, `libx264` often just falls back to 25fps anyway.

**Setup**
 - OS: Linux
 - Shell: bash
 - Terminal Emulator: N/A (VHS rendering)
 - Terminal Multiplexer: N/A

**To Reproduce**
1. Create a tape file with `Output demo.mp4` and `Set Framerate 50` (or 60).
2. Run `vhs demo.tape`.
3. Check the framerate of the resulting video using `ffprobe demo.mp4`.
4. See that the video is `25 fps` instead of 50.

**Source Code**
```tape
Output demo.mp4
Set Framerate 50
Set Margin 20
Set MarginFill "#000000"
Type "echo hello"
Enter
Sleep 1s

shoemoney:
The window bar and both margin-fill sources default to 25fps and sit in the *main* slot of their overlays, so ffmpeg clamps the whole render to 25fps no matter what `Set Framerate` says.

Measured on a 50fps tape: 25fps/62 frames before, 50fps/123 frames after, with duration unchanged. Half of every captured frame was being discarded silently.

MarginFill defaults to the theme background, so this affects every render, not only tapes that set a margin.

Fixes #750

- [x] I have created a discussion that was approved by a maintainer (for new features).

### DIFF
--- a/ffmpeg.go
+++ b/ffmpeg.go
@@ -11,6 +11,7 @@ import (
 type FilterComplexBuilder struct {
 	filterComplex *strings.Builder
 	style         *StyleOptions
+	framerate     int
 	termWidth     int
 	termHeight    int
 	prevStageName string
@@ -50,6 +51,7 @@ func NewVideoFilterBuilder(videoOpts *VideoOptions) *FilterComplexBuilder {
 		termHeight:    termHeight,
 		termWidth:     termWidth,
 		style:         videoOpts.Style,
+		framerate:     videoOpts.Framerate,
 		prevStageName: "padded",
 	}
 }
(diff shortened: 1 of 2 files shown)


shoemoney:
Green against main and rebase-clean. Honors Set Framerate on overlay input streams. Ready for review.

naqa92:
+1

PR #788, opened 2026-09-11
fix: preserve context for rendering and report encoder errors

### LINKED ISSUE #787: v0.12.0 exits 0 and writes no output file on ubuntu-24.04 runners; v0.11.0 records the same tape
On an `ubuntu-24.04` GitHub Actions runner, vhs v0.12.0 runs a tape to completion, prints `Creating <file>.gif...`, prints the publish hint, exits 0, and writes no file. v0.11.0 records the same tape on the same runner without complaint.

The gap between the two printed lines is about 200 to 440 milliseconds, which is far too short to encode twenty seconds of terminal, so it looks like the encode never starts rather than failing.

## Controlled comparison

Two runs of the same job, on the same runner image, from two commits whose only functional difference is the version passed to `charmbracelet/vhs-action@v2.1.1`:

| Run | vhs | Result |
|---|---|---|
| [34459564668](https://github.com/vahapogut/trustdiff/actions/runs/34459564668) | v0.12.0 | exits 0, no file anywhere on the filesystem |
| [34460025896](https://github.com/vahapogut/trustdiff/actions/runs/34460025896) | v0.11.0 | writes a 201 KB, 1200x700, 224 frame GIF |

The diff between the two commits is one line plus comments:

```diff
-          version: v0.12.0
+          version: v0.11.0
```

Both runs are public, so the full logs are readable.

## Environment

```
vhs     v0.12.0 (db96d73)   /opt/hostedtoolcache/vhs/0.12.0/x64/...
ttyd    1.7.7-40e79c7
ffmpeg  6.1.1-3ubuntu5
chrome  Google Chrome 152.0.7977.64  (chromium 152.0.7977.0 also present)
os      Ubuntu 24.04.5 LTS, x86_64
image   ubuntu-24.04, version 20260831.293.1
action  charmbracelet/vhs-action v2.1.1
```

## Tape

```
Output docs/demo.gif

Require trustdiff

Set Shell bash
Set FontSize 16
Set Width 1200
Set Height 700
Set Padding 20
Set TypingSpeed 40ms

Type "trustdiff check npm:express@4.19.2 pypi:requests cargo:serde"
Sleep 500ms
Enter
Wait+Screen@300s /Exit code/
Sleep 6s
```

`trustdiff` is a CLI on PATH; any command that prints a screen of text and ends with a line matching the `Wait+Screen` pattern should do.

## Observed

Every tape step executes and every `Wait+Screen` matches, so the terminal side is working. Then:

```
09:16:20.0019515Z  Creating docs/demo.gif...
09:16:20.0021686Z
09:16:20.4461187Z  Host your GIF on vhs.charm.sh: vhs publish <file>.gif
```

`vhs docs/demo.tape` run by hand in the same job exits 0. Afterwards `docs/` contains no gif, and `find / -name 'demo.gif' -type f -newermt '-5 minutes'` finds nothing.

## Ruled out

Four things, each tested on its own before the version was suspected:

- **`Set Columns` / `Set Rows`.** The tape originally used them, and they are new in v0.12.0, so they were the obvious suspect. Replacing them with `Set Width` / `Set Height` changed nothing.
- **Missing ffmpeg.** The `ubuntu-24.04` image does not carry ffmpeg. Installing it changed nothing here, but see the note below.
- **A `Hide` / `Show` block.** The tape warmed a cache off camera inside one. Removing it entirely changed nothing.
- **Tape length.** Cut from two recorded commands to one. Changed nothing.

## Possibly a separate issue

vhs does not appear to check for ffmpeg before starting, and a run without it fails in exactly the shape above: tape completes, `Creating ....gif...` prints, exit 0, no file. That made this bug considerably harder to isolate, because the first thing anyone checks produces the same symptom. A startup check, or an error when the encode produces nothing, would turn both cases into something a reader can act on. Happy to open that separately if you would rather keep them apart.

Thanks for vhs. The recording it produces is exactly what the project wanted.

gudcks0305:
Fixes #787.

Recording teardown canceled the context subsequently passed to ffmpeg, so completed tapes could exit successfully without creating their output. Keep recording cancellation on a child context and use the caller's context for rendering. Return encoder errors instead of only logging their output, so a missing ffmpeg executable or a failed render reaches the caller.

Added a regression test for missing ffmpeg and an opt-in browser test that records a short tape, decodes the resulting GIF, and checks cancellation immediately before rendering. All three cases failed on the original code and pass with the fix.

Validation on macOS arm64:
- `go test -race ./...`
- `go vet ./...`
- `VHS_TEST_BROWSER=1 go test -run 'Test(RenderMissingFFmpeg|EvaluateGIF)$' -count=1 -v .` (real Chrome, ttyd and ffmpeg)

Implemented with OpenAI Codex assistance and independently reviewed with Codex before submission.

### DIFF
--- a/vhs.go
+++ b/vhs.go
@@ -568,6 +568,7 @@ func (vhs *VHS) Render(ctx context.Context) error {
 		out, err := cmd.CombinedOutput()
 		if err != nil {
 			log.Println(string(out))
+			return fmt.Errorf("render failed: %w", err)
 		}
 	}
 
--- a/evaluator.go
+++ b/evaluator.go
@@ -113,8 +113,8 @@ func Evaluate(ctx context.Context, tape string, out io.Writer, opts ...Evaluator
 	}
 
 	// Begin recording frames as we are now in a recording state.
-	ctx, cancel := context.WithCancel(ctx)
-	ch := v.Record(ctx)
+	recordCtx, cancel := context.WithCancel(ctx)
+	ch := v.Record(recordCtx)
 
 	// Clean up temporary files at the end.
 	defer func() {
--- a/evaluator_test.go
+++ b/evaluator_test.go
@@ -1,0 +1,68 @@
+package main
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	imagegif "image/gif"
+	"io"
+	"os"
+	"os/exec"
+	"path/filepath"
+	"testing"
+	"time"
+)
+
+func TestRenderMissingFFmpeg(t *testing.T) {
+	t.Setenv("PATH", t.TempDir())
+	v := New()
+	v.totalFrames = 1
+	t.Cleanup(func() { _ = v.Cleanup() })
+	v.Options.Video.Output.GIF = filepath.Join(t.TempDir(), "output.gif")
+
+	if err := v.Render(context.Background()); !errors.Is(err, exec.ErrNotFound) {
+		t.Fatalf("expected missing ffmpeg error, got %v", err)
+	}
+}
+
+func TestEvaluateGIF(t *testing.T) {
+	if os.Getenv("VHS_TEST_BROWSER") == "" {
+		t.Skip("set VHS_TEST_BROWSER=1 to run this test")
+	}
+	for _, name := range []string{"ttyd", "ffmpeg"} {
+		if _, err := exec.LookPath(name); err != nil {
+			t.Fatalf("%s is required: %v", name, err)
+		}
+	}
+
+	for _, cancelRender := range []bool{false, true} {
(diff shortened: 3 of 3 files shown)


vhs.go
```diff
@@ -568,6 +568,7 @@ func (vhs *VHS) Render(ctx context.Context) error {
 		out, err := cmd.CombinedOutput()
 		if err != nil {
 			log.Println(string(out))
+			return fmt.Errorf("render failed: %w", err)
```
andrinoff: Returning on the first failed encoder means if the GIF fails, the MP4/WebM/screenshots for the same tape are never attempted. Consider collecting the errors with errors.Join and running all commands

andrinoff (approved):
thank you!

